PR #42, opened 2019-01-09
Checkbox → Switch in /host/listings

woeltjen:
## Description
Display the "Publish" control as a switch instead of a checkbox, which is more consistent with the original design documents.

Also adds a `title` to display instructions to complete listing on hover when this control is disabled

## How to Test
Verify aesthetics and check for regressions relative to instructions on #26 

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
The messaging is fairly ad hoc; we may want to revise in the future.

There are likely to be merge conflicts with #40, so one of these PRs will need conflict resolution depending on which we merge first.

## Learnings
Many React components [expect use of `css-loader`](https://github.com/gfazioli/react-toggle#default-styles) or similar


src/components/routes/Host/HostListingCard/HostListingCard.tsx
```diff
@@ -16,6 +16,8 @@ interface Props extends HostListingShort {
   deactivateListing: (id: string) => Promise<Listing>;
 }
 
+const INCOMPLETE_LISTING = "This listing is incomplete. Use the Edit button to complete this listing to publish.";
```
tommy-bee: Maybe add something to the effect of all fields must be filled to publishable?

src/components/routes/Host/HostListingCard/HostListingCard.container.ts
```diff
@@ -30,13 +30,22 @@ const HostListingCardContainer = styled.article`
...
       label {
-        margin: 7px;
+        display: inline-flex;
+        max-height: 40px;
+        padding: 6px;
+        span {
+          ${typography('caption', 2)};
+          padding: 6px;
+        }
+        span.host-listing-meta--disabled {
```
kcvan: small but i think you can do something like:
```
span {
  &.host-listing-meta--disabled {
  }
}

to stay consistent with scss/sass

**Comment by jerminatorhits** [comment]
When I edit an `isActive` listing, then click `Save & Exit`, the listing becomes `inActive`.

I believe two things need adjusting:
- remove `isActive` attribute from `defaultValues` in `ListingForm.tsx` so when the form submits, it doesn't send `isActive: true`.

- backend currently has this code in `createListing` and `updateListing`
```
isActive: user.isAdmin() ? isActive : false,
```
removing it should allow hosts to set `isActive`

src/components/routes/Host/HostListingCard/HostListingCard.tsx
woeltjen: Sure, worked in the phrase "all required fields" (there are a couple of optional fields, so "all fields" seemed misleading)

tc: 👍

kcvan:
functionality looks good! is it possible to change the color of the green to 'correct'? LGTM besides that!

woeltjen:
@kcvan Switched to `correct` as the color; looks much nicer, thanks for pointing that out!

`react-switch` only accepts hex-strings for colors; `color` from `styled/utils` returns `rgba(....)` colors. Added a `hexColor` export to work around this, but let me know if you see a cleaner way

PR #45, opened 2019-01-10
create NumberInput for beds (review)

jerminatorhits:
## Description
Why did you write this code?
https://beetoken.atlassian.net/browse/ENG-842

- Build a reusable component that always has a `number` value (in native input fields, users can just 'backspace' to delete it). Use it for `numberOfBedrooms`, `numberOfBathrooms`, and `maxGuests` so far.
- Change `sleepingArrangement` naming
- Swap `sharedBathroom` input component with a `Checkbox`

Allows us to scale with accommodating different bed types (especially beneficial as we integrate with the bigger fish)

## How to Test
- [x] Visit `host/listings/:id/accommodations` and `.../pricing_availability`
- [x] Click `+` and `-` and see changes saved accordingly.
- [x] Click checkbox and see 'Yes' or 'No' in the db corresponding to the checkbox value.

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

Reused `Fab` component in this one. Found out `Fab` needs updating. I wanted to make the svg icons bigger so it's easier to click but it was difficult to add the dynamic prop with it's outdated architecture -- will make a separate ticket for this.

## Learnings


src/components/shared/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+interface State {
+  value: number;
+}
+
+class NumberInput extends React.Component<Props, State> {
+  static defaultProps: Props = {
+    value: 0,
+    max: Number.MAX_SAFE_INTEGER,
+    min: Number.MIN_SAFE_INTEGER,
```
tc: should the default min be `0` instead?

jerminatorhits: ya `0` makes sense for practical purposes~

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
```diff
@@ -31,57 +32,61 @@ const AccommodationsForm = (props: any): JSX.Element => {
...
+          <InputLabel>Number of Bedrooms</InputLabel>
+          <NumberInput
+            value={values.numberOfBedrooms}
+            max={50}
+            min={1}
+            onChange={(value: number) => {
+              setFieldValue('numberOfBedrooms', value);
+              setFieldTouched('numberOfBedrooms');
+            }}
+          />
```
tc: does this no longer need an error message because the user can't select a bad input?

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
```diff
@@ -31,57 +32,61 @@ const AccommodationsForm = (props: any): JSX.Element => {
...
+              setFieldTouched('numberOfBathrooms');
+            }}
+            step={0.5}
+          />
         </div>
       </div>
 
+      <div className="form-item">
+        <Checkbox
+          checked={values.sharedBathroom === 'Yes' || values.sharedBathroom === 'yes'}
```
tc: you can call `toUpperCase()` to normalize the string comp.

tc: why don't we use a boolean for this instead of a string?

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
jerminatorhits: ya 👍

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
jerminatorhits: boolean would be ideal for `sharedBathroom`. I think it was a string from the beginning =X

jerminatorhits: will do. also will include other wonky responses ppl have put

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
```diff
@@ -98,3 +103,14 @@ const AccommodationsForm = (props: any): JSX.Element => {
...
+// Temporary fix until sharedBathroom is changed into a boolean
+function isSharedBathroom(input: string): boolean {
+  if (!input) return false;
+  if (parseInt(input) && (parseInt(input) !== 0)) {
+    return true;
+  }
+  if (input.length > 10) return true;
+  const normalizedInput = input.toLowerCase();
+  return normalizedInput === 'yes';
+}
```
jerminatorhits: ran scripts on prod data to test:

image

woeltjen: The Lower Bedroom Bath is sort of a shared because Liviing room and Kitchen is down there  : true

🤣

This looks reasonable as a temporary fix. The code didn't make sense to me until I saw the examples from production; maybe amend the comment above with a note like "this is specific to values in existing listings" or similar?

Also (I'm just code golfing here) I think this is equivalent to:

    return !!input && (parseInt(input) > 0 || input.length > 10 || input.toLowerCase() === 'yes');

tc: Can we manually replace `The lower bathroom is sort of....` with a ``true`` or ``yes`` in the database? I think it would be better if this was a `yes` and the disclaimer be left in the description or other house notes.

src/components/shared/NumberInput/NumberInput.container.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+interface Props {
+  font?: string;
+  textColor?: string;
+  end?: string;
+  start?: string;
+}
+
+const typographyMap: ValueMap = {
```
kcvan: did we decide to move these out to a utils file or somewhere where we can reuse it?

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
kcvan: should we manually add it into the house notes or make a note to the host about it?

src/components/shared/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,115 @@
...
+          disabled={value === min}
+          height="24px"
+          icon="utils/minus-circle-outline"
+          iconColor={prefixColor || 'secondary'}
+          noPadding
+          onClick={this.handleMinus}
+          type="button"
+          width="24px"
+        />
+        <span>{this.state.value}</span>
```
woeltjen: Should this wrap an [ `<input type='number'>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) instead of just a span?

From the issue description, I see that there is an issue where users can clear the contents. Could we address that by wrapping an `input` here? (e.g. `onChange` check if text is empty/non-numeric and restore prior value when so)

Using `<input type='number'>` gives us:

* Accessibility (ability to tab in/out, semantics for screen readers)
* Flexibility (user can type a number instead of just clicking buttons)
* Numeric keyboards (on mobile)
* More stuff that I'm forgetting (probably)

src/components/shared/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,115 @@
...
+  value: number;
+  max: number;
+  min: number;
+  onChange: (value: number) => void;
+  step: number;
+
+  end?: string;
+  font?: string;
+  prefixColor?: string;
+  prefixSize?: string;
```
kcvan: i think this, suffixSize, and svgSize arent being implemented atm?

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
jerminatorhits: i manually changed `The lower bathroom...` to `Yes` in the prod db.

src/components/shared/NumberInput/NumberInput.container.ts
jerminatorhits: talked in person. will make a separate ticket to condense these (this particular one is being used in 4 places)

src/components/shared/NumberInput/NumberInput.tsx
jerminatorhits: removed ty

src/components/shared/NumberInput/NumberInput.tsx
jerminatorhits: tried a few css tricks with `input`. Found out that it has a default `width` which makes it difficult to conditionally render css `padding` between svgs in the long run.

Talked with @kcvan pointed out that users don't actually need to tab into the number itself -- tabbing on airbnb just selects the buttons and they just use `div`s and `button`s.

image

PR #55, opened 2019-01-12
Improved autocomplete suggestions

jerminatorhits:
## Description
Why did you write this code?
swap react-google-maps SearchBox with self google api component since the library doesn't allow in-depth customizations for location

## How to Test
- [x] Visit '/' or dev03.beenest.io
- [x] Type inside input and see cities suggested
- [x] Verify all query params are still valid and being sent on submit

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

`react-google-maps` doesn't expose very many options to customize as beenest grows. It's good for small/side projects so as our product grows, it may be good to slowly move away.

### DIFF
--- a/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
+++ b/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
@@ -35,8 +35,7 @@ class GoogleAutoComplete extends React.Component<Props, any> {
   }
 
   handlePlaceChanged = () => {
-    const a = this.autocomplete.getPlace();
-    this.props.onPlaceChange(a);
+    this.props.onPlaceChange(this.autocomplete.getPlace());
   }
 
   render() {
--- a/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
+++ b/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 
 import { compose, withProps } from 'recompose';
 import { withScriptjs } from 'react-google-maps';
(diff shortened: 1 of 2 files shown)


src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
```diff
@@ -11,30 +10,50 @@ import { SETTINGS } from 'configs/settings';
...
+  autocomplete: AutocompleteInterface;
+  
+  componentDidMount() {
+    if (!this.props.inputRef.current) return;
 
-class StandaloneSearchBoxWithErrorBoundary extends React.Component<BoundaryProps> {
-  state = { hasError: false };
+    this.autocomplete = new google.maps.places.Autocomplete(
+      this.props.inputRef.current,
+      {"types": ["(cities)"]}
```
tc: does using the google maps api directly let us prioritize recommended cities? ie, typing in `new` should show `new york` 

I did notice on dev03 typing in `new` showed `newark, new york` vs some random city on beenest.com

jerminatorhits: I think only these types can be used https://developers.google.com/places/supported_types#table3

I'll do a bit more research to see it can be adjusted further. If it can't be changed, another autocomplete library could be an option.

tc: we should support cities (new york) and regions (hawaii)

tc:
@kcvan +1 it's in a better state than we found it.

PR #58, opened 2019-01-14
add undefined check

jerminatorhits:
## Description
Why did you write this code?
https://sentry.io/beetoken/beenest/issues/844681455/?referrer=slack

Error occurs when users types into autocomplete field and presses 'Enter' key without clicking the suggested options.

Clicking the suggested options will get the coordinates, but for the case mentioned above, this prevents the app from trying to get 'unfetched' coordinates.

## How to Test
- [ ] Visit /
- [ ] Open console 
- [ ] type in 'los' or 'new york' and press enter, verify error doesn't show anymore

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/shared/SearchBar/SearchBar.tsx
+++ b/src/components/shared/SearchBar/SearchBar.tsx
@@ -156,6 +156,8 @@ class SearchBar extends React.Component<RouterProps, State> {
   handleGuests = (event: React.ChangeEvent<HTMLInputElement>) => this.setState({ numberOfGuests: event.target.value });
   
   handlePlaceChange = (place: google.maps.places.PlaceResult) => {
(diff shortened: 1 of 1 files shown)


tc:
if i press enter, lat/lng of 0 shouldn't be submitted:
 ``http://localhost:4200/listings?locationQuery=san&utm_term=san&coordinates[lat]=0&coordinates[lng]=0&numberOfGuests=1``

PR #60, opened 2019-01-15
Integrate token swaps at time of payment

woeltjen:
## Description
This integrates Uniswap (via the [Unipay](https://github.com/woeltjen/unipay) "Payment Agent" contract) into the payment flow to allow BEE payments which originate in DAI.

Tracking to-dos:

- [x] Show DAI as an option
- [x] Get pricing at run-time
- [x] Call Uniswap during payment
- [x] Provide messaging during payment steps
- [x] Add margin for precision, price fluctuation
- [x] Figure out why guestConfirmBooking isn't being called...
- [x] Make sure guestConfirmBooking advances booking state correctly
- [x] Handle case when liquidity is just unavailable
- [x] Verify that `invoice` works
- [x] Fix test failures
- [x] Code cleanup? (Itemize future refactors?)

## How to Test
1. Get some "Ropsten DAI" (really just an ad hoc ERC-20 token)
    1. You can `mint` all you like on [etherscan](https://ropsten.etherscan.io/address/0x909e4dbdef114c9b39078df314177780b89d8062#writeContract)
2. Go make a booking
    1. There is limited Ropsten BEE liquidity; prefer cheap listings (e.g. actual nest of bees)
    2. (Or, find a very expensive booking and verify that payment is disabled with useful messaging...)
3. Select "DAI" as a payment option
4. Click through to pay
    1. Confirm two MetaMask transactions
5. Go to `/admin` to confirm the bookings; verify `invoice` call completes.

- Chrome on Mac
## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. More payment types
2. Backend initiation of the `collect` call

## Learnings
The [`using...for`](https://solidity.readthedocs.io/en/v0.5.2/contracts.html#using-for) keyword in Solidity allows contract-scoped decoration of arbitrary types; this is useful when you want contract interactions to be both [diligently-checked](https://github.com/woeltjen/unipay/blob/5c99fdebbe360efdc7a8bd887e6c9bdd1935c6f6/contracts/Unipay.sol#L179-L241) and yet [concise](https://github.com/woeltjen/unipay/blob/5c99fdebbe360efdc7a8bd887e6c9bdd1935c6f6/contracts/Unipay.sol#L283-L286)


src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
```diff
@@ -45,6 +49,7 @@ class SelectPaymentOption extends React.Component<Props> {
                   Select Payment Type
                 </option>
                 {showBee && <option value={Currency.BEE}>BEE</option>}
+                {showBee && <option value={Currency.DAI}>DAI</option>}
```
kcvan: small comment, but how do we want to order the select box? from the options we want them to select the most to least?

woeltjen: Good question! I'm not sure additional ERC-20 currencies belong in this select box at all, particularly as we grow to accept more. In lieu of trying to tackle that design problem I'm trying to fit it in with minimal deviation from the existing UI.

For now, I've ordered it alphabetically and welcome any other guidance or direction.

woeltjen:
Currently blocked by issues making calls in sequence, which is necessary for the approve-and-collect flow; filed MetaMask/metamask-extension#6028

Investigating workarounds. Could move the `collect` call to the admin side (into an agent which calls `collect` then `invoice`) but that is going to be far more sensitive to price volatility. May just need to have some UI to initiate the second call?

woeltjen:
Tested an [`invoice`](https://ropsten.etherscan.io/tx/0x638bec01d22979f27b4072c29f17591e61e4dba5d68581ee454432b48942c1df) call, and that works! Manually provided `guestWalletAddress` and `hostWalletAddress` on the booking (edited `meta` in the database) in order to work around incomplete payment flow due to the above; so, at minimum still need to verify that these fields are populated appropriately on `guestConfirmBooking` when that can be reached.

woeltjen:
Notes on code cleanup:

- `fromBee` is an awkward way to communicate between components
  - Especially when used to communicate that exchange is unavailable
- Wrapping PriceQuote works but ended up bleeding up to the interface
- Pricing lookup (`priceWithToken` call) seems a bit awkwardly located?
- Async in BookingPayment leads to some really awkward nesting

Going to remove WIP status on these, since all could reasonably allocated to a separate refactor. In particular, with BTC and channel manager integration upcoming, it looks like future payment methods will be integrated here, and it would be useful too consider those requirements in a refactor.

woeltjen:
Marking WIP again; need to do a mainnet unipay deployment

config/settings.js
```diff
@@ -26,6 +26,10 @@ module.exports = {
     STRIPE_CLIENT_KEY: 'pk_test_7GNkTJHiPbROv4gPDuQNvThc',
     STRIPE_CLIENT_ID: 'ca_CmC1grl4bBfjYWWwKg548UPazsb6USUW',
     SENTRY_CLIENT_DSN: 'https://ee59845fce1746e4a2d6d664fd7a441b@sentry.io/1193097',
+    TOKEN_ADDRESSES: {
```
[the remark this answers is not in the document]
woeltjen: Yep. What about `ERC20_ADDRESSES`?

src/components/routes/Booking/BookingPayment/BookingPayment.tsx
```diff
@@ -32,47 +39,66 @@ const BookingPayment = ({ history, match }: RouterProps) => (
...
-                      >
-                        Go Back
-                      </Button>
-                      <BookingPaymentButton
-                        booking={booking}
-                        onSuccess={() => history.push(`/bookings/${booking.id}/receipt`)}
-                      />
+        <Async promise={pricePromise} then={price => {
+          const fromBee = (!!price && currency !== booking.currency) ?
+            ((value: number) => value * price * 1.01 / booking.guestTotalAmount) :
```
tc: 1.01 ? is this error margin for approval?

config/settings.js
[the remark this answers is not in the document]
tc: +1

src/components/routes/Booking/BookingOptions/BookingOptionsCrypto/BookingOptionsCrypto.tsx
```diff
@@ -51,85 +53,92 @@ const BookingOptionsCrypto = ({ booking, currency, history }: Props) => (
...
-            </AppConsumer>
-          </BookingOptionsCryptoContainer>
+                  );
+                }}
+              </AppConsumer>
+            </BookingOptionsCryptoContainer>
+          );
+        };
+        return (
+          <Async promise={renderContents()} then={contents => contents} />
```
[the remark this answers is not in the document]
tc: im not sure i understand this Async component usage.

You are saying we need to wait on:
```
  const availableFunds = await getAvailableAmount(accounts, currency); 
```
Is this needed before the component loads?
If so, could this have happened in ``didComponentMount``?

Is `accounts` or `currency` being changed from a higher component? if so, could the prop just be changed and this component be re-rendered?

Alternatively, would something like HOC take care of the currency concerns and pass down to this rendered component? https://reactjs.org/docs/higher-order-components.html

src/components/routes/Booking/BookingPayment/BookingPayment.tsx
woeltjen: Yep, there are two cases where this matters:

1. When values in Uniswap exchanges change between price being queried and user submitting payment
2. When there is a loss of precision going back and forth to the JS `number` type

The second case could be addressed by keeping prices as strings in wei/dust, but that seemed like an uphill battle (much of the current code assumes numbers) and we'd still need to do something similar for the first case.

src/components/routes/Booking/BookingOptions/BookingOptionsCrypto/BookingOptionsCrypto.tsx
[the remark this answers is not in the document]
woeltjen: > Is this needed before the component loads?

It's needed before I can display the user's available DAI. The case is different from BEE/ETH, because those values are available as soon as Web3Provider provides `accounts`, but that pattern won't scale going forward (we'd need to query balances on every ERC-20 coin we support prior to passing `accounts` to the consumer)

> If so, could this have happened in `didComponentMount`?

Only with a larger refactor. Retrieving the balance depends on the address from `accounts`, which will not be available until after render.

> Is `accounts` or `currency` being changed from a higher component? if so, could the prop just be changed and this component be re-rendered?

`accounts` comes from the usage of Web3Provider in this component; `currency` comes from props.

> Alternatively, would something like HOC take care of the currency concerns and pass down to this rendered component?

Applying a HOC to the whole component would run into the same issue as `componentDidMount`: Don't have `accounts` yet.

---

One option here would be to change the usage of Async here to retain more of the original structure:

    <Async promise={getAvailableAmount(accounts, currency)} then={availableFunds => {
      const total = fromBee ? fromBee(quote.guestTotalAmount) : quote.guestTotalAmount;
      // ...etc...
    }} />

Or, I could just do that larger refactor and split this out into parts so we can introduce a HOC/`componentDidMount` at the appropriate level.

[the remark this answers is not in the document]
kcvan: the `<Async>` component is such a powerful package...but IT'S SO SIMPLE.  Why didn't we think of this😅

kcvan:
has a method to test this been determined?😈

src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
```diff
@@ -67,18 +79,29 @@ class SelectPaymentOption extends React.Component<Props> {
...
 }
 
 export default SelectPaymentOption;
 
-function currencyOptions(currency: string | undefined, booking: Booking): React.ReactNode {
+function currencyOptions(currency: string | undefined, booking: Booking, fromBee?: (value: number) => (number | string)): React.ReactNode {
   switch (currency) {
     case Currency.BEE:
-      return <BookingOptionsCrypto booking={booking} currency={Currency.BEE} />;
+    case Currency.DAI:
```
jerminatorhits: do the BEE and DAI cases need to return anything?

woeltjen: I'm using [fallthrough](https://en.wikipedia.org/wiki/Switch_statement#Fallthrough) here (BEE/DAI/ETH all return the same component)

woeltjen:
@kcvan Thanks for the reminder, added test instructions

src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
jerminatorhits: interesting! TIL

src/components/routes/Booking/BookingOptions/BookingOptionsCrypto/BookingOptionsCrypto.tsx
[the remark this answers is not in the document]
tc: i think ``<Async promise={getAvailableAmount(accounts, currency)}`` is more understandable what you are doing than ``<Async promise={renderContents()} then={contents``

src/components/routes/Booking/BookingPayment/BookingPaymentButton.tsx
```diff
@@ -21,14 +23,16 @@ interface State {
   isSubmitting: boolean;
 }
 
+const FROM_BEE_MESSAGE = 'You will be prompted to Confirm two Ethereum transactions to complete this payment.';
```
tc: Would this alert the user more?
``You will be prompted to Confirm two Ethereum transactions`` => 
``You will be prompted to confirm TWO Ethereum transactions`` ?

PR #65, opened 2019-01-17
Trips Tabs

jerminatorhits:
## Description
Why did you write this code?
Make a set of tabs on Trips page to filter guest bookings
Uses: https://github.com/thebeetoken/beenest-backend/pull/654

## How to Test
- [x] Visit /trips
- [x] Click nav items
- [x] Verify `/profile` navbar doesn't break

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?
Yes.

- Some stylings are repeated -- will need to rethink / rearchitect the css classes to make the implementation cleaner.
- Also will need to brute force test long values to see if the design holds up with verbose details.

## Learnings


src/components/routes/Account/AccountNav/AccountNav.container.ts
```diff
@@ -12,6 +12,7 @@ const AccountNavContainerMobile = styled.nav`
     background-color: ${color('light')};
     color: ${color('secondary')};
     display: flex;
+    flex-basis: 0;
```
tc: is this the CSS just for the tab component?

we should avoid specific names like `verification-needed` so we can just re-use it as different use cases instead of copying and pasting.
it can be a generic name like`alert-badge`

cc @kcvan

src/components/routes/Account/AccountNav/AccountNav.tsx
```diff
@@ -1,93 +1,60 @@
...
                 }
-        
+
                 return (
-                  <BeeLink to="/account/verification" isNav activeClassName="active">
-                    Verification
+                  <BeeLink to={to} isNav activeClassName="active" key={to}>
+                    {src
+                      ? <Svg src={src} />
+                      : <h2>{title}</h2>
```
tc: are h2 appropriate for tab links?

src/components/routes/Trips/Trips.tsx
```diff
@@ -76,60 +76,97 @@ class Trips extends React.Component<Props, State> {
...
+                      to: '/trips/past',
+                    },
+                    {
+                      title: 'Cancelled/\nRejected',
+                      to: '/trips/cancelled',
+                    }
+                  ]} />
+                  <div className="trip-cards-container">
+                    <Switch>
+                      <Route exact path="/trips/current" component={() =>
```
tc: when visiting /trips, if there are current trips, show current trip, else show upcoming.

The upcoming tab should have an empty state like "Book your next travel now"

kcvan:
this is back to WIP, right? @jerminatorhits

jerminatorhits:
ya

src/components/routes/Trips/Trips.tsx
jerminatorhits: updated to use the empty default: 

image

src/components/shared/TabNav/TabNav.tsx
```diff
@@ -11,13 +11,27 @@ interface Props {
...
 }
 
 
 const TabNav = (props: Props): JSX.Element => {
-  const renderTabNavItems = props.config.map(({ title, to }: Config) => {
+  const renderTabNavItems = props.config.map(({ showBadge, title, to }: Config) => {
+    if (showBadge) {
+      return (
+        <div className="verification-needed-container">
+          <span className="verification-badge">!</span>
```
woeltjen: Would we be able to get rid of the `showBadge` block and instead move the `verification-badge` span into the same place below? It would deduplicate some code and clarify what the condition really applies to

    {showBadge && <span className="verification-badge">!</span>}

src/components/routes/Trips/StartedTripCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Listing } from 'networking/listings';
+
+interface Props {
+  trip: Booking;
+}
+
+const StartedTripCard = ({ trip }: Props) => {
+  const { checkInDate, checkOutDate, listing, numberOfGuests } = trip;
+  const { idSlug } = listing;
+  const streetAddress = (listing.addressLine1 || '').concat(listing.addressLine2 ? `, ${listing.addressLine2}` : '');
```
woeltjen: Is StartedTripCard for bookings where `status` is `started`? If so, I don't you'll have the address fields unless you're an admin user

src/components/routes/Trips/StartedTripCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+      </div>
+    </StartedTripCardContainer>
+  );
+};
+
+export default StartedTripCard;
+
+function getGoogleMapURI(listing: Listing): string {
+  const listingAddress = `${listing.addressLine1 ? listing.addressLine1 + ' ' : ''}${listing.city} ${listing.state}`;
+  return `https://www.google.com/maps/place/${encodeURIComponent(listingAddress)}`;
```
woeltjen: Corollary to the above, I'd use `lat`/`lng` from the listing here, since an anonymized version of these will be available for `started` bookings

jerminatorhits: using `lat/lng` now

src/components/shared/TabNav/TabNav.tsx
jerminatorhits: yea that would be ideal. I can conditionally render the span, but I'm having trouble finding ways to conditionally render parent div:

```
<div className="verification-needed-container">
     <span className="verification-badge">!</span>
     {children}
</div>
````
Maybe ternary might be a tad clearer?

Comment (woeltjen) [reply src/components/shared/TabNav/TabNav.tsx]: Does the parent `div` need conditional rendering? It seems like it could be a neutral container when the `span` isn't there, but I haven't really looked at CSS

If CSS does matter, you could try this on the parent:

    <div className={showBadge ? 'verification-needed-container' : ''}>

**Comment by woeltjen** [comment]
Currently getting this when I click on Trips on this branch:

image

File: src/components/shared/TabNav/TabNav.tsx (hunk 4 above)
Comment (jerminatorhits) [reply src/components/shared/TabNav/TabNav.tsx]: good suggestion; I'll try to make the container more neutral and will comment when implemented!

Comment (jerminatorhits) [reply src/components/shared/TabNav/TabNav.tsx]: updated

File: src/components/routes/Account/AccountNav/AccountNav.container.ts (hunk 1 above)
Comment (jerminatorhits) [reply src/components/routes/Account/AccountNav/AccountNav.container.ts]: changed `verification-badge` to `alert-badge`
changed `<AccountNav />` to `<ReusableNavBar />`

File: src/components/routes/Account/Account.tsx [hunk 7]
```diff
@@ -15,6 +14,7 @@ import AudioLoading from 'shared/loading/AudioLoading';
 import Divider from 'shared/Divider';
 import GeneralWrapper from 'shared/GeneralWrapper';
 import NotFound from 'routes/NotFound';
+import ReusableNavBar from 'shared/ReusableNavBar';
```
tc: maybe just `TabNavBar`? a component by definition should be `reusable`

src/components/routes/Trips/ActiveTripCard.container.ts
```diff
@@ -102,14 +97,17 @@ const ActiveTripCardDesktop = styled(ActiveTripCardTablet)`
         }
       }
       .dates {
-        height: 18px;
-        width: 324px;
-        h5 {
-          ${typography('read', 2)};
-          span {
+        .date {
+          h5 {
```
tc: a bigger problem i see with our CSS setup is that we have to alter the default tags like `h5`

Typically a project should define sensible defaults for its styles and a programmer can just use `p, h1` tags at will without adding more CSS.
see https://getbootstrap.com/docs/4.2/content/typography/

src/components/routes/Trips/ActiveTripCard.tsx
```diff
@@ -24,23 +23,24 @@ interface Props {
...
       </div>
-      <div className="trip-large-section">
+      <div className="active-trip-info">
         <h3>
           <BeeLink to={titleLink}>{listing.title}</BeeLink>
         </h3>
         <div className="bee-flex-div" />
         <div className="address">
-        <BeeLink href={getGoogleMapURI(listing)} target="_blank">
+        <BeeLink href={`https://www.google.com/maps/search/?api=1&query=${lat},${lng}`} target="_blank">
```
tc: i think keeping the `getGoogleMapURI` method is sensible and just changing the address to lat/lng.

src/components/routes/Trips/ActiveTripCard.tsx
```diff
@@ -24,23 +23,24 @@ interface Props {
...
           <ListItem noHover suffixColor="secondary" textColor="secondary" textTransform="uppercase">
             <span>{trip.status === 'host_approved' && streetAddress} {listing.city && `${listing.city}, `}{listing.state && `${listing.state}, `}{listing.country.toUpperCase()}</span>
```
tc: don't we have a address formatter here?

src/components/routes/Account/Account.tsx
jerminatorhits: renamed

src/components/routes/Trips/StartedTripCard.container.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    .address {
+      margin-bottom: 16px;
+      .bee-list-item {
+        span {
+          ${typography('emp', 7)}
+        }
+      }
+    }
+    .dates {
```
tc: any reason to redefine dates css?
shouldn't this be the same css as other date css in cards?

src/components/routes/Trips/ActiveTripCard.container.ts
jerminatorhits: that sounds a lot cleaner. Maybe we can slowly transition to using default mappings? @kcvan

src/components/routes/Trips/Trips.tsx
tc: 👍

src/components/routes/Trips/Trips.tsx
```diff
@@ -59,77 +60,111 @@ class Trips extends React.Component<Props, State> {
...
+                        </>
+                      } />
+                      <Route exact path="/trips/upcoming" component={() =>
+                        <>
+                          {isUpcomingEmpty
+                            ?
+                              <div className="trips-book-now">
+                                <div className="trips-book-now--text">
+                                  <h2>You haven't booked any trips yet.</h2>
+                                  <BeeLink to="/">Let's change that!</BeeLink>
```
tc: i dont think we need two links, maybe get rid of the Let's change that link and keep the button?

src/components/routes/Trips/ActiveTripCard.tsx
jerminatorhits: reinstated `getGoogleMapURI` and put it in `utils/formatter` since it is used in both `StartedTripCard` and `ActiveTripCard`

src/components/routes/Trips/ActiveTripCard.tsx
jerminatorhits: using `formatAddress` now

src/components/routes/Trips/StartedTripCard.container.ts
jerminatorhits: currently we are defining the styles explicitly for each `StartedTripCard`, `ActiveTripCard`, and `ExpiredTripCard`. Should we make a reusable "CardInfo" component that has the styles?

tc: yeah, i think they are the same styles?

src/components/routes/Trips/Trips.container.ts
```diff
@@ -2,16 +2,24 @@ import styled from 'styled-components';
...
-  justify-content: center;
   align-items: center;
+  display: flex;
+  min-height: 100vh;
+  width: 100%;
 
   
-  .trips-wrapper {
+  .trips-body {
+    align-items: flex-start;
```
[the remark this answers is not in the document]
jerminatorhits: ty removed

src/components/routes/Trips/Trips.tsx
jerminatorhits: sg. removed `Let's change that!`

src/utils/formatDate.ts
```diff
@@ -16,8 +16,8 @@ export const formatDateRange = (checkInDate: string, checkOutDate: string): stri
  * @param {string} string ex. 2019-01-09T00:00:00.000Z
  * @returns {string} ex. 01/09/2019
  */
-export const formatSingleDate = (date: string): string => {
-  return format(date.slice(0, 10), 'MM/DD/YY');
+export const formatSingleDate = (date: string, formatString?: string): string => {
+  return format(date.slice(0, 10), formatString || 'MM/DD/YYYY');
```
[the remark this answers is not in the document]
jerminatorhits: date-fns `format` takes in 3 args: (date, format, options). `locale` can be passed in as an options.

I think the locales need to be manually imported in order to implement them.
https://date-fns.org/v2.0.0-alpha.7/docs/I18n

src/components/routes/Trips/StartedTripCard.container.ts
jerminatorhits: the sketch file has minor variations such as the dates being placed in different locations, the google maps icon exists for Active Cards, and different fabs depending on the state (this is already being done in our existing cards with condition rendering, but will be more logic in one reusable component). Should we consolidate into one component and forgo the small design differences? If so, I'm thinking I may need a props flag for cardSize: `big` and `small`.

image

tc: i see, the difference is the black "checkin date" and the  "12/01/2017" in the small card.
i think its fine to keep the difference then.

PR #66, opened 2019-01-17
add referral code

tc:
## Description
Add Referral code

## How to Test
- [ ] Visit /hosts/signup?code=TOMTOM
- [ ] Signup
- [ ] Check https://app.segment.com/beenest-development/sources/node-js/debugger that an event was triggered


kcvan:
It looks like the referral code only worked when it was in the url for me.  I tried url only (worked) and promo code in form only (didnt work)

tc:
@kcvan can u retry? I took out the input field and added a display. kieran mentioned it can reduce conversion.

kcvan:
LGTM! one small thing, can we change the text to match the other input titles?

PR #67, opened 2019-01-17
Use lat/lng bounds when searching

woeltjen:
## Description
Get the lat/lng boundary associated with a search term from Google's APIs, and pass that information along to the back-end so that queries for larger regions (e.g. "Hawaii") don't fail due to the search area chosen by the backend.

Note that this depends on thebeetoken/beenest-backend#649 and should be merged after that PR has been merged (and not before)

## How to Test
1. Search "Hawaii"
2. Verify that results appear

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/components/routes/Listing/ListingsResult/ListingsResult.tsx
+++ b/src/components/routes/Listing/ListingsResult/ListingsResult.tsx
@@ -54,10 +54,12 @@ const ListingQuery = () => {
     numberOfGuests: numberOfGuests ? parseInt(numberOfGuests) : 1,
     locationQuery: locationQuery || '',
     ...(areBoundsValid && {
-      east: parseFloat(bounds.east),
-      north: parseFloat(bounds.north),
-      south: parseFloat(bounds.south),
-      west: parseFloat(bounds.west),
+      bounds: {
+        east: parseFloat(bounds.east),
+        north: parseFloat(bounds.north),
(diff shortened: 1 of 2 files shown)


kcvan:
Great update, Vic!  Really refines (...and expands?😂) our searches!

works, LGTM!

kcvan:
One thing Jeremy and I noticed though (nothing to do with your pr, just the google search in general), when we type in `hawaii` by itself, nothing shows up.  But if we click on the dropdown that says `hawaii, usa` it works.  I wonder if there's a way to make it work just for `hawaii`?  I'm sure a lot of users just type in the city or state and expect results to show

woeltjen:
> One thing Jeremy and I noticed though (nothing to do with your pr, just the google search in general), when we type in `hawaii` by itself, nothing shows up.  But if we click on the dropdown that says `hawaii, usa` it works.  I wonder if there's a way to make it work just for `hawaii`?  I'm sure a lot of users just type in the city or state and expect results to show

Thanks! Filed thebeetoken/beenest-backend#653 to handle this case.

PR #76, opened 2019-01-22
Restrict contract cancellation to times when guest has paid

woeltjen:
## Description
The "Cancel" button on the "Trips" page was causing MetaMask to show in cases when the guest had not actually been charged yet. This restricts the contract call to cases where the guest has actually paid.

## How to Test
1. Make a new BEE booking
    1. Note that this must be >1wk in the future for the original bug to be reproducible
2. Do *not* accept as a host
3. Go to Trips
4. Cancel
5. Verify that MetaMask does *not* appear

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/components/routes/Trips/Trips.tsx
+++ b/src/components/routes/Trips/Trips.tsx
@@ -173,9 +173,9 @@ export default compose(
   graphql(GUEST_CANCEL_BOOKING, {
     props: ({ mutate }: any) => ({
       cancelBooking: async (booking: Booking) => {
-        const { id, currency, checkInDate } = booking;
+        const { id, currency, checkInDate, status } = booking;
         const days = differenceInDays(checkInDate, Date.now());
-        if (currency === Currency.BEE && days >= 7) {
+        if (currency === Currency.BEE && days >= 7 && status === 'guest_paid') {
           const web3 = loadWeb3();
           await cancel(web3.eth, id);
         }


src/components/routes/Trips/Trips.tsx
```diff
@@ -173,9 +173,9 @@ export default compose(
   graphql(GUEST_CANCEL_BOOKING, {
     props: ({ mutate }: any) => ({
       cancelBooking: async (booking: Booking) => {
-        const { id, currency, checkInDate } = booking;
+        const { id, currency, checkInDate, status } = booking;
         const days = differenceInDays(checkInDate, Date.now());
-        if (currency === Currency.BEE && days >= 7) {
+        if (currency === Currency.BEE && days >= 7 && status === 'guest_paid') {
```
jerminatorhits: would this restrict the `cancel` to only 'guest_paid'? I'm not too familiar with the metamask flow but would we want to allow `cancel` for other statuses further in the process like 'host_approved' as well?

woeltjen: One thing to note is that this allows cancelling the booking regardless of status (or, it obeys whatever status constraints have been specified elsewhere); the call being conditioned here is the call to the contract to reclaim funds.

Looking over the statuses defined in the Booking model:

                'started',
                'expired_before_guest_confirmed',
                'guest_confirmed',
                'guest_cancel_initiated',
                'guest_cancelled',
                'guest_rejected_payment',
                'host_rejected',
                'host_approved',
                'host_cancelled',
                'expired_before_host_approved',
                'init_pay_submitted',
                'payment_failed',
                'guest_paid',
                'host_paid',
                'dispute_initiated',
                'disputed',
                'refund_initiated',
                'refunded',
                'completed'

There is a case where status is `host_approved` but the backend hasn't yet picked up on the event to advance the status to `guest_paid`. This is an ambiguous situation; we don't know if the host's `invoice` call actually went through (if we did, status would be `guest_paid`), so we don't know if the Payments contract has any funds for the user to reclaim.

I think that erring on the side of caution here means looking at just `guest_paid`, insofar as (In the event that a user does find a way to hit that edge case) we can handle the refund with an off-line process.

woeltjen: (More broadly, all of this is why I want to decouple booking and payment status 😁)

jerminatorhits: ahh i see. so it practically only needs to 'refund' the contract when `guest_paid` since the money actually left the guest wallet at that point. sounds good!

+1 on decoupling the statuses!

PR #79, opened 2019-01-22
Add BTC to payment options

woeltjen:
## Description
Adds BTC as a payment option to increase reach to BTC hodlers

## How to Test
1. Find a listing, start a booking
2. Select BTC as payment option
3. Click forward through "Submit Payment"
4. Expect to see a Bitcoin wallet address with the amount due and instructions for payment

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
This implementation will be a simple address that guests can send BTC to. Will we want to automate this? We'll see!

## Learnings
Bees communicate the location of interesting food sources by flying in figure eights and wiggling their butts, and so do I.


woeltjen:
image

tc:
I think we need some text to the effect of "If the host does not accept the trip, the payment amount will be returned to the sender address."

PR #84, opened 2019-01-24
deploy

tc:
btc work, firebase handler(not linked from email yet)


tc:
tested search, booking with CC. works

PR #88, opened 2019-01-25
sharedBathroom default 'No'. Allow click 'Save & Continue' while form is pristine

jerminatorhits:
## Description
Why did you write this code?
sharedBathroom field required is stifling users from publishing listings.
https://beetoken.atlassian.net/browse/ENG-917

**Not sure if we should make a backend change instead / as well. There are listings with `sharedBathroom: ''` which could be empty string forever if we remove it as a required.

## How to Test
- [x] Create a listing and publish without touching sharedBathroom.

- [x] Click `Save & Continue` without touching the form and the app should proceed successfully

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/routes/Host/ListingForm/ListingForm.tsx
+++ b/src/components/routes/Host/ListingForm/ListingForm.tsx
@@ -48,7 +48,7 @@ const defaultValues: FormValues = {
   postalCode: '',
   pricePerNightUsd: 100,
   securityDepositUsd: 50,
-  sharedBathroom: '',
+  sharedBathroom: 'No',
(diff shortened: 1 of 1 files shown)


tc:
lgtm, as discussed, further work can be to make this a boolean

PR #93, opened 2019-01-28
make unpublish notice more clear

tc:
## Description
We got over 15 support emails wondering why publish is disabled.

## How to Test
- [ ] Visit /host/listings
- [ ] See text:
screen shot 2019-01-28 at 10 59 13 am

## Learnings
Tooltips are too hidden

### DIFF
--- a/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
+++ b/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
@@ -53,6 +53,7 @@ const HostListingCardContainer = styled.article`
 
   .host-listing-notice {
     ${typography('caption', 3)};
+    max-width: 125px;
   }
 
   .host-listing-image {
--- a/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
+++ b/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
@@ -48,8 +48,12 @@ const HostListingCardContainer = styled.article`
         }
       }
     }
+
   }
 
+   .host-listing-notice {
+    ${typography('caption', 3)};
+   }
 
   .host-listing-image {
     height: 100%;
--- a/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
+++ b/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
@@ -51,9 +51,9 @@ const HostListingCardContainer = styled.article`
 
   }
 
-   .host-listing-notice {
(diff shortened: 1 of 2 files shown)


src/components/routes/Host/HostListingCard/HostListingCard.container.ts
```diff
@@ -48,13 +48,17 @@ const HostListingCardContainer = styled.article`
         }
       }
     }
+
   }
 
+   .host-listing-notice {
+    ${typography('caption', 3)};
```
woeltjen: Inconsistent whitespace here (one space instead of two)

src/components/routes/Host/HostListingCard/HostListingCard.tsx
```diff
@@ -62,11 +62,14 @@ const HostListingCard = (props: Props): JSX.Element => {
...
           <Button background="core" color="white" size="small" onClick={() => deleteListing(id)}>
             Delete
-          </Button>          
+          </Button>
+          <div className='host-listing-publish'>
           <label htmlFor={`publish-${id}`} title={canPublish ? '' : INCOMPLETE_LISTING}>
             <span className={canPublish ? '' : 'host-listing-meta--disabled'}>Publish</span>
             <Switch checked={isActive} disabled={!canPublish} onColor={hexColor('correct')} onChange={() => toggleListing(id)} id={`publish-${id}`} />
           </label>
+          {!canPublish && <p className='host-listing-notice'>{INCOMPLETE_LISTING}</p>}
```
woeltjen: Whitespace here too; expect `<label ...>` and `{!canPublish...}` to be indented

woeltjen:
From the screenshot, it looks like the text is causing that last item to grow wide enough to push the buttons together. Maybe a `max-width` on `.host-listing-notice`?

src/components/routes/Host/HostListingCard/HostListingCard.tsx
tc: 👍  why cant lint find this i wonder

jerminatorhits:
+1 to vic's comment.

I think the text would work with this design:
image

tc:
@jerminatorhits we'll impl jasmine's new design in a future PR.

I added a max-width so the buttons look normal now.

jerminatorhits:
getting a leak on long titles. Is it possible to shrink the msg to 3 lines?

image

tc:
sounds good, shortened text.

PR #95, opened 2019-01-28
add basic errors on listing form

jerminatorhits:
## Description
Why did you write this code?
add errors to form.

WIP Update: At first I had required params based on PUBLISH_PROPERTIES. By having fields `.required()` on the Yup schema, when a user edits just the title and clicks save & continue, the form gets mad and the errors show up under the required fields (even if you didn't touch them). This doesn't seem user friendly since it will force hosts to fill out the info that they might not have ready yet. On the flip side, it might lead to more listings created faster.

Now this current implementation removes the `.required()` so there is loose validation such as if the title is too short, too long, etc. and the intention is to decouple 'filling out the listing info' and 'being ready to publish'. When a listing is to be published, then we can tell our users of missing required attributes. Suggestions about this are welcome!

WIP UPDATE 2: Found that click another tab before saving causes the 'required' errors to show. Currently finding a workaround.

## How to Test
- [ ] Visit `host/listings/` or dev03.beenest.io and click a create/edit a listing.
- [ ] Test invalid inputs on each field.

Verify the required fields correlate the ones in the backend:
```
const PUBLISH_PROPERTIES = [
  'checkInTime',
  'checkOutTime',
  'country',
  'title',
  'description',
  'pricePerNightUsd',
  'listingPicUrl',
  'addressLine1',
  'postalCode',
  'sleepingArrangement',
  'sharedBathroom',
  'minimumNights',
  'maxGuests',
  'photos',
  'amenities',
  'houseRules'
];
```

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?
Make the states fade-in / fade-out

Possibly place the errors to the right of the inputs if design requires it.

## Learnings


woeltjen:
When I try to navigate from Listing Info to Accommodations, I see the following prompt to confirm:

image

Not sure where this prompt is being generated, but at minimum I'd expect that array to get turned into a string (`errors.join('\n')` somewhere?)

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
```diff
@@ -41,6 +42,9 @@ const ListingInfoForm = (props: FormikProps<ListingInput>): JSX.Element => {
             <Svg className="suffix" src="utils/carat-down" />
           </label>
         </SelectBoxWrapper>
+        <ErrorMessageWrapper>
+          <ErrorMessage name="homeType" />
+        </ErrorMessageWrapper>
```
woeltjen: Coding in an `ErrorMessageWrapper` for each form input creates a lot of repetition. Can this be hidden behind `InputWrapper` and similar classes? (Or should we punt that until we decide to implement BeeForm?)

jerminatorhits: ah `<ErrorMessage />` is a `Formik` component. It renders a simple string with no divs. `<ErrorMessageWrapper />` was used to tag the component and provide css.

`<ErrorMessage />` can also render its own divs. Maybe we can leverage the small api provided to clean up our implementation. https://jaredpalmer.com/formik/docs/api/errormessage. I'll keep thinking/trying different implementations throughout this week =D

src/components/routes/Host/ListingForm/ListingForm.tsx
```diff
@@ -60,9 +60,9 @@ interface Props extends RouterProps {
 }
 
 const ListingFormSchema = Yup.object().shape({
-  addressLine1: Yup.string().min(1, 'Too Short!'),
+  addressLine1: Yup.string().required('Address Line 1 is a required field'),
```
kcvan: can we move these error messages to a utils folder for easy access/altering?

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
```diff
@@ -41,6 +42,9 @@ const ListingInfoForm = (props: FormikProps<ListingInput>): JSX.Element => {
             <Svg className="suffix" src="utils/carat-down" />
           </label>
         </SelectBoxWrapper>
+        <ErrorMessageWrapper>
```
kcvan: for the errors and inputs, do you think there's a way to loop through and dynamically create these?

src/components/routes/Host/ListingForm/ListingForm.tsx
jerminatorhits: just made a `formatError` function in `utils/formatter`. apollo gql also have a `formatError` function; hope the imports won't be confusing with the same function name

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
jerminatorhits: cool thanks! updated

jerminatorhits:
@woeltjen The alert should look prettier now!

tc:
screen shot 2019-01-29 at 4 23 19 pm

The date error looks unreadable.
Should we throw up a modal instead of an alert to make it more readable?

src/components/routes/Host/ListingForm/ListingForm.container.ts
```diff
@@ -37,20 +37,23 @@ const ListingFormContainer = styled.section`
...
-            display: flex;
-            flex-direction: row;
-            justify-content: space-between;
-            .bee-input-wrapper {
-              &:nth-of-type(1) {
-                width: 45%;
-              }
-              &:nth-of-type(2) {
+          >.bee-input-wrapper {
+            &:nth-of-type(3) {
```
tc: what are these nth type for?  is it for the ordering style?
if it's for targeting certain fields like `title`, they should be attached via class names since fields can change order.

jerminatorhits: 3, 4, and 5 are for the `city`, `state`, and `postalCode` styles respectively:

image

I could attach classNames instead but attributes like the `margin-left` would be constrained to that input so if a dev switched the order, the margins would be off. Originally had a div that wrapper the three with class `row-address` that took care of the styles/spacing in that row.

@kcvan suggested this to `map` through an array of objects to render the address inputs: https://github.com/thebeetoken/beenest-web/pull/95#discussion_r251677561

tc: can we add these attributes to a class that lives at the `bee-input-wrapper` level? that would make the `nth-of-type` css sensible. otherwise a programmer wouldn't know why 
```
>.bee-input-wrapper {
            &:nth-of-type(3) {
``` 
is related to an address formatting.

jerminatorhits:
Alert now looks like this:
image

Further down the line, we could make util functions that prettify client errors and server side errors, assuming they are sent in a predictable format.

src/components/routes/Host/ListingForm/ListingForm.container.ts
jerminatorhits: added `.city`, `.state`, `.postal-code` within `.address`

tc:
Can the Len be expanded to length on title is too short

kcvan:
can we map out all these validation errors or is there a place we can find all of them? would be nice to document them and be ready for any errors that pop up

jerminatorhits:
@kcvan Figured I would need to make multiple custom errors for each field (such as min, max, and missing) which would be a huge object (unless we just make a set of generic messages).

Spent a bit of time trying to make Yup errors human readable; they have ways to modify the keys themselves to make it readable but that ruins formik networking. They also have their own string interpolation that can't be short circuited. For example `'${value} is too long.'` this is passed into the schema as a string verbatim and spits out as `<lowercamelcased key> is too long.` https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema

Still figuring out the best way to organize and maintain the error messages but I think this will work for now

jerminatorhits:
error message should be formatted better now!

As for test examples, most (if not all) of the string inputs just ask for a least 1 char. The number input errors can be tested with negative numbers; the other ones use the `NumberInput` component which should be free of errors. ical urls can be invalid urls.

With the addition of `.required()` on several fields I'm currently running into a problem where when I click `Save & Continue`, the error list includes the fields that are on other tabs since I'm sending the entire listing form every time. May need to break out the form into 4 small baby forms after all in order to leverage the desired error behavior. I believe formik links its `errors`/`isValid` and its `onSubmit` so it's difficult to allows users to submit with errors present.

Starting to 'Build forms in React, **with** the tears 😭'

jerminatorhits:
UPDATE: Was able to bypass `Formik` errors and just 'submit with errors' to the backend using a declared `handleSubmit`

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
```diff
@@ -7,7 +7,8 @@ import Textarea from 'shared/Textarea';
 import { TextareaEvent } from 'shared/Textarea/Textarea';
 import { stringToArray, arrayToString } from 'utils/formatter';
 import Checkbox from 'shared/Checkbox';
-import { Field } from 'formik';
+import { Field, ErrorMessage } from 'formik';
+import ErrorMessageWrapper from 'components/shared/ErrorMessageWrapper';
```
kcvan: `components` snuck into here

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
```diff
@@ -33,15 +34,16 @@ const AccommodationsForm = (props: any): JSX.Element => {
       </div> */}
```
kcvan: are we keeping the commented code for later?

src/components/routes/Host/ListingForm/ListingForm.container.ts
```diff
@@ -37,21 +37,22 @@ const ListingFormContainer = styled.section`
...
-            }
+          .city {
+            display: inline-block;
+            margin-bottom: 0;
+            width: 45%;
+          }
+          .state {
+            display: inline-block;
+            margin-bottom: 0;
+            margin-left: 53px;
```
kcvan: i think these margin lefts should be `10%` each to fill out the remaining space

src/components/routes/Host/ListingForm/ListingForm.container.ts
```diff
@@ -37,21 +37,22 @@ const ListingFormContainer = styled.section`
...
-                width: 45%;
-              }
-              &:nth-of-type(2) {
-              width: 15%;
-              }
-              &:nth-of-type(3) {
-                width: 20%;
-              }
-            }
+          .city {
```
kcvan: small thing but i think you can do:
```
.city,
.state,
.postal-code {
  display: inline-block;
  margin-bottom: 0;
}
```

to stop some redundancy

src/components/routes/Host/ListingForm/ListingForm.tsx
```diff
@@ -199,10 +205,7 @@ class ListingForm extends React.Component<Props, State> {
 
                   <Button
                     onClick={() => {
-                      if (!FormikProps.isValid) {
-                        alert(`Cannot save changes due to errors: ${JSON.stringify(Object.values(FormikProps.errors), null, 4)}`);
-                      }
-                      FormikProps.submitForm();
+                      this.handleSubmit(FormikProps.values, FormikProps);
                     }}
                     textStyle="title-8"
```
kcvan: i think title-8 is default

src/components/routes/Host/ListingForm/ListingForm.tsx
```diff
@@ -220,6 +223,23 @@ class ListingForm extends React.Component<Props, State> {
...
+    const { props } = this;
+    actions.setSubmitting(true);
+    const { updateListing } = props;
+    const { id } = props.match.params;
+    return updateListing(id, values)
+      .then(() => {
+        props.history.push(`/host/listings/${this.state.nextCrumb}`);
+      })
+      .catch((error: ApolloError) => {
+        const formattedError = error.graphQLErrors ? error.graphQLErrors.map(e => e.message).join('\n').toString() : error;
```
kcvan: i think `/n` is normally for unix/max osx +
https://stackoverflow.com/questions/15433188/r-n-r-and-n-what-is-the-difference-between-them/15433225#15433225

you might want to use a combination of `/n` and `/r` to hit windows

src/components/shared/ErrorMessageWrapper/ErrorMessageWrapper.container.ts
```diff
@@ -3,9 +3,8 @@ import { color, typography } from 'styled/utils';
 
 const ErrorMessageWrapperContainer = styled.div`
   ${typography('light', 7)}
-  color: ${color('incorrect')};
+  color: ${color('error')};
   height: 32px;
-  width: 100%;
   transition: all 0.1s ease-in-out;
```
kcvan: i think this might not do anything atm?

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
jerminatorhits: 👍

src/components/routes/Host/ListingForm/AccommodationsForm/AccommodationsForm.tsx
jerminatorhits: yea. However, I'm not sure how high pri adding additional fields is to our existing system and with the shifting business needs.

src/components/routes/Host/ListingForm/CheckinDetailsForm/CheckinDetailsForm.tsx
```diff
@@ -6,6 +6,8 @@ import Svg from 'shared/Svg';
 import Textarea from 'shared/Textarea';
 import { TextareaEvent } from 'components/shared/Textarea/Textarea';
 import timeOptions from 'utils/timeOptions';
+import ErrorMessageWrapper from 'components/shared/ErrorMessageWrapper';
```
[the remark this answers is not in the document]
jerminatorhits: 👍

src/components/routes/Host/ListingForm/ListingForm.container.ts
jerminatorhits: mah man 👍

src/components/routes/Host/ListingForm/ListingForm.container.ts
jerminatorhits: updated

src/components/routes/Host/ListingForm/ListingForm.tsx
jerminatorhits: tyty

src/components/shared/ErrorMessageWrapper/ErrorMessageWrapper.container.ts
jerminatorhits: removed

jerminatorhits:
In order to allow submissions with errors present, I'll be adding backend validations as well

jerminatorhits:
Form now lets users submit. Added a "(required)" label to inputs that need a value for publishing. Some might confuse it with "required for submission" vs "required for publishing" but I think it'll provide overall positive guidelines.

Form should still work as intended. Let me know if any further issues arise

src/components/routes/Host/ListingForm/ListingForm.tsx
```diff
@@ -220,6 +211,23 @@ class ListingForm extends React.Component<Props, State> {
...
+    actions.setSubmitting(true);
+    const { updateListing } = props;
+    const { id } = props.match.params;
+    return updateListing(id, values)
+      .then(() => {
+        props.history.push(`/host/listings/${this.state.nextCrumb}`);
+      })
+      .catch((error: ApolloError) => {
+        const formattedError = error.graphQLErrors ? error.graphQLErrors.map(e => e.message).join('\r\n').toString() : error;
+        alert(`${formattedError}\r\n\r\nIf this continues to occur, please contact us at support@beetoken.com`);
```
tc: support at beenest

src/components/routes/Host/ListingForm/ListingFormNav/ListingFormNav.tsx
```diff
@@ -63,18 +62,18 @@ const ListingFormNav = ({ errors, history, id, isValid, onSubmit, setNextCrumb,
       <GeneralWrapper width={976}>
         <Prompt
           when={showAlert}
-          message={`Listing has unsaved changes ${!isValid ? `due to the following errors: ${JSON.stringify(Object.values(errors), null, 4)}`: ''}. Are you sure you want to proceed?`}>
+          message={`Listing has unsaved changes ${!formikProps.isValid ? `due to the following errors:\n\n${Object.values(formikProps.errors).join('\n').toString()}\n\n`: ''} Are you sure you want to proceed?`}>
```
tc: can we put this into a function? ``formatListingErrorsAlert(formikProps.errors)`` or something similar?

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
```diff
@@ -125,46 +141,74 @@ const ListingInfoForm = (props: FormikProps<ListingInput>): JSX.Element => {
...
+        </ErrorMessageWrapper>
       </div>
 
       <div className="form-item map-preview">
         <InputLabel>Location Preview on Map</InputLabel>
         <GoogleMaps
           address={address}
           getCoordinates={({ lat, lng }) => {
-            if ((lat || lat === 0) && (values.lat || values.lat === 0) && Math.abs(lat - values.lat) > LAT_LNG_EPSILON) {
+            if (
```
tc: is this the same as below? can you assign it to a function? 
``isCoordinateValid(lat)``

src/components/routes/Host/ListingForm/ListingFormNav/ListingFormNav.tsx
jerminatorhits: made new function on bottom

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
jerminatorhits: refactored out 👍

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
tc: Is there an instance where <ErrorMessage /> is used with <ErrorMessageWrapper /> ?
you can write a short helper:
```
const StyledErrorMessage = (prop) => <ErrorMessageWrapper>
<ErrorMessage {...prop}>
</ErrorMessageWrapper>;
```

tc:
i edited a listing, removed the title, clicked save, it showed the error, then it instantly went to an white screen error. This is the console log:

screen shot 2019-02-01 at 10 04 24 am

jerminatorhits:
fml. I wrote `formikProps.error` instead of `formikProps.errors` 😭

tc:
```
ERROR in [at-loader] ./src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx:154:35
    TS2345: Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.

ERROR in [at-loader] ./src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx:157:35
    TS2345: Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.
ℹ ｢wdm｣: Failed to compile.
```

src/components/routes/Host/ListingForm/ListingInfoForm/ListingInfoForm.tsx
jerminatorhits: updated

jerminatorhits:
made `isCoodinateValid` accept `number | undefined` instead of just `number`

PR #97, opened 2019-01-29
Handle case where host is undefined

woeltjen:
...to unbreak Create Host, which uses the same form but provides no host object with which to initialize it.

## Description
Sentry raised some errors on `/admin/users/new`; these were caused by changes from #92, which utilize `listingCount` and `id` of a host. These properties are not present when creating a host, so a fallback is needed.

## How to Test
1. Navigate to `/admin/users`
2. Click Create Host
3. Verify that you see the "Create User Profile" form instead of crashing the app

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
What else did I break? 🙃

## Learnings
Test editing *and* creating when a change touches both.

### DIFF
--- a/src/components/Admin/AdminUsers/AdminUsersHostForm/AdminUsersHostForm.tsx
+++ b/src/components/Admin/AdminUsers/AdminUsersHostForm/AdminUsersHostForm.tsx
@@ -285,7 +285,7 @@ class AdminUsersHostForm extends React.Component<HostProps, HostFormState> {
   render() {
     const { errors } = this.state;
     const { about, btcWalletAddress, email, firstName, lastName, phoneNumber, profilePicUrl, walletAddress } = errors;
-    const { id, listingCount } = this.props.host;
+    const { id, listingCount } = this.props.host || { id: undefined, listingCount: undefined };
     const nameError = firstName.error || lastName.error;
     const nameSuccess = firstName.success && lastName.success;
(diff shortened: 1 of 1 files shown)


src/components/Admin/AdminUsers/AdminUsersHostForm/AdminUsersHostForm.tsx
```diff
@@ -285,7 +285,7 @@ class AdminUsersHostForm extends React.Component<HostProps, HostFormState> {
   render() {
     const { errors } = this.state;
     const { about, btcWalletAddress, email, firstName, lastName, phoneNumber, profilePicUrl, walletAddress } = errors;
-    const { id, listingCount } = this.props.host;
+    const { id, listingCount } = this.props.host || { id: undefined, listingCount: undefined };
```
kcvan: if the host is `undefined` will it just show `undefined` as the id and listingCount in the admin portal?

jerminatorhits: ^there's conditional rendering at the bottom

PR #101, opened 2019-01-30
Allow same-day date selection

woeltjen:
## Description
Don't block out two days in advance in the date picker for booking requests. This allows guests to make bookings on the day of stay.

## How to Test
1. View a listing without any bookings today
2. Verify that the current day is shown in the date picker
3. Proceed to book with check-in set to today
4. Verify that crypto options are not shown
5. Verify that USD payment completes as expected
6. View a new listing
7. Select a day more than 48 hours in advance, proceed to book
8. Verify that crypto payment options *do* show

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
We may want to change our approach to crypto payments to work with this

## Learnings
I hope time zones turn out to be a passing fad.

### DIFF
--- a/src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
+++ b/src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import moment from 'moment';
 
 import { Booking, Currency } from 'networking/bookings';
 
@@ -14,6 +15,8 @@ import SelectBoxWrapper from 'shared/SelectBoxWrapper';
 import Svg from 'shared/Svg';
 import { AppEnv, APP_ENV } from 'configs/settings';
 import { loadWeb3, priceWithToken } from 'utils/web3';
+
+const TWO_DAYS_MS = 2 * 24 * 60 * 60 * 1000;
 
 interface Props {
   booking: Booking;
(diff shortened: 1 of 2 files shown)


woeltjen:
Note that this also needs some checks when selecting payment option; BEE payments will fail this near in the future, with the current contract.

kcvan:
when clicking next page where we select the payment option, a alert popped up asking to select a payment option when it was already on credit card.  cant reproduce though and everything else went through

woeltjen:
> when clicking next page where we select the payment option, a alert popped up asking to select a payment option when it was already on credit card. cant reproduce though and everything else went through

Was the error "There was an issue. Please contact us at support@beetoken.com!" by any chance? I'm not able to reproduce either, so I'm seeing if I can find the root cause by analysis. (The easiest way to reproduce that particular error is to kill the backend before clicking the button.)

kcvan:
the error that popped up was 'please select a payment option', which is a little strange because it should default to credit card,  this might be related to @jerminatorhits fix on auto selecting a payment option and credit card

woeltjen:
@kcvan Ah, thanks, looks like [this error](https://github.com/thebeetoken/beenest-web/blob/ccf5098a06fd35717c0d0b9cead22cae39b30a70/src/components/routes/Booking/BookingPayment/BookingPayment.tsx#L38-L41), which implies that `booking.currency` isn't being set. Hmm...

kcvan:
could have also been a fluke, i havent been able to repro so 🎉

kcvan:
`I hope time zones turn out to be a passing fad.` dying

woeltjen:
> could have also been a fluke, i havent been able to repro so 🎉

Still unable to repro here, too. Looking at code paths, I see three ways for `booking.currency` to remain undefined when we reach BookingPayment:

1. The `writeQuery` to update the cache for the booking hasn't completed and the cached result of `GET_BOOKING` (from before `currency` selection) is being used
2. `guestSelectPayment` isn't updating the `currency` property on the backend in some unpredictable way.
3. `guestSelectPayment` *is* updating `currency` on the backend, but returning a stale version of the object.

That said, I'm not able to get any of those three things to occur.

My thinking is that there is a bug in payment selection, but that it is hard to reproduce, and unlikely to have been introduced in this PR. So, inclined to defer further investigation until we have better reproduction steps. (If any reviewers see a way to trigger this bug from this changeset, that inclination will change.)

PR #106, opened 2019-01-31
Show pricing card when screen type is tablet

woeltjen:
## Description
During booking, the pricing card would disappear when selecting payment options with a "tablet" screen type, but the bottom bar would not appear until the screen type becomes "mobile"

This change shows the card when screen type is "tablet" or larger, consistent with what is displayed in subsequent booking steps.

## How to Test
1. Make a new booking
2. Play with the size of the window
3. Expect the bottom bar to appear whenever the pricing card disappears

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
+++ b/src/components/routes/Booking/BookingOptions/SelectPaymentOption/SelectPaymentOption.tsx
@@ -77,7 +77,7 @@ class SelectPaymentOption extends React.Component<Props> {
         </div>
         <AppConsumer>
           {({ screenType }: AppConsumerProps) =>
-            screenType > ScreenType.TABLET && (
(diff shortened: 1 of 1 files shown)


jerminatorhits:
This is a good fix for now.

It'd be nice to have some padding/margin on the sides but the components themselves are too big to allow any: 
image

Ideally I think the intended behavior was for the bottom bar to show up on TABLET as it does on the listing page, but it might take some complex css z-indexing since the 'X' to close is covered. (just tried it real quick, the `BookingOptionsBar` seems to be nested in quite a few layers.)

image

PR #107, opened 2019-01-31
More Calendar link in host listings card

woeltjen:
## Description
Moves the Calendar link on a host's listing card from a button to text, per updated designs

Note that the updated design doesn't include "Last edited" text, so I made some judgment calls there

## How to Test
1. Go to `/host/listings`
2. Verify that "View Calendar >" shows under title and metadata of each listing
3. Verify that link still works

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/components/routes/Host/HostListingCard/HostListingCard.tsx
+++ b/src/components/routes/Host/HostListingCard/HostListingCard.tsx
@@ -8,6 +8,7 @@ import HostListingCardContainer from './HostListingCard.container';
 import BeeLink from 'shared/BeeLink';
 import Button from 'shared/Button';
 import LazyImage from 'shared/LazyImage';
+import Svg from 'shared/Svg';
 import { ACTIVATE_LISTING, DEACTIVATE_LISTING, DELETE_LISTING, GET_HOST_LISTINGS, HostListingShort, Listing } from 'networking/listings';
 import { formatAddress } from 'utils/formatter';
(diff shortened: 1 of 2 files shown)


woeltjen:
image

src/components/routes/Host/HostListingCard/HostListingCard.tsx
```diff
@@ -41,20 +42,20 @@ const HostListingCard = (props: Props): JSX.Element => {
     <HostListingCardContainer className="host-listing-card">
       <div className="host-listing-meta">
         <h1>{title}</h1>
-        {(city || state || country) && <h2>{formatAddress(city, state, country)}</h2>}
+        <h2>{(city || state || country) ? formatAddress(city, state, country) : ''}</h2>
         <div className="bee-flex-div" />
         <h3>Last edited: {format(updatedAt, 'MM/DD/YY [at] hh:mmA')}</h3>
+        <h4>
```
kcvan: i think you can do a button with the `clear` and `suffix` properties to achieve something similar

woeltjen: Thanks, I'll look at that

PR #114, opened 2019-01-31
Deploy (revert revert)

woeltjen:
Reverts thebeetoken/beenest-web#112 (redeploy GraphQL changes now that the backend is deploying)


tc:
@woeltjen  can we  merge this into master and then into production?

PR #122, opened 2019-02-01
firstName/lastName → displayName

woeltjen:
## Description
Uses changes from thebeetoken/beenest-backend#711

## How to Test
This is (mostly) a non-functional change; regression test listings where host information is shown.

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Removal of deprecated fields on the backend

### DIFF
--- a/src/components/Admin/AdminListings/AdminListingsTable/AdminListingTableRow.tsx
+++ b/src/components/Admin/AdminListings/AdminListingsTable/AdminListingTableRow.tsx
@@ -36,7 +36,7 @@ class AdminListingTableRow extends React.Component<Props> {
         </td>
         <td className="admin-table-row--item">
           <BeeLink to={host ? `/admin/users/${host.id}` : ''}>
-            <span>{host ? `${host.displayName}` : ''}</span>
+            <span>{host ? `${host.fullName}` : ''}</span>
           </BeeLink>
           <CopyToClipboard text={host ? host.email : 'email does not exist'}>
             <span>{host && host.email ? host.email : 'email does not exist'}</span>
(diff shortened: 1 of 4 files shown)


src/components/Admin/AdminListings/AdminListingsTable/AdminListingTableRow.tsx
```diff
@@ -36,7 +36,7 @@ class AdminListingTableRow extends React.Component<Props> {
         </td>
         <td className="admin-table-row--item">
           <BeeLink to={host ? `/admin/users/${host.id}` : ''}>
-            <span>{host ? `${host.firstName} ${host.lastName}` : ''}</span>
+            <span>{host ? `${host.displayName}` : ''}</span>
```
woeltjen: This part is actually a change, at least right now; the backend returns `displayName` as `firstName`

We currently hide `lastName` from non-admins at the GraphQL layer. To do something analogous here, we could add some backend logic to compute `displayName` as `${firstName lastName}` for admins and `firstName` for everyone else.

That said, I'm wondering if that information is useful enough here for us to care about that

woeltjen: Also, I should note that I haven't tested this change (blocked by another issue, filed ENG-954)

woeltjen: Sorry, I think my description was confusing there.

The line of code we are discussing introduces the regression you describe; previously, we showed `firstName` and `lastName` in the admin listings table, with this change we do not.

In the context of the other PR, I realized I could add a [`fullName`](https://github.com/thebeetoken/beenest-backend/pull/711/files#r253246385) property with admin-only permissions to avoid this regression. Think I'll go ahead and do that.

kcvan: works! thanks vic

PR #123, opened 2019-02-02
Kevin/ENG-928

kcvan:
## Description
Why did you write this code?
accommodate for rentivo bookings through trips receipt

to be used with: https://github.com/thebeetoken/beenest-backend/pull/709
## How to Test
- [ ] create a rentivo booking in thailand I(not mine)
- [ ] confirm that the trips receipt page looks good

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/routes/Trips/TripsReceipt/TripsReceipt.tsx
+++ b/src/components/routes/Trips/TripsReceipt/TripsReceipt.tsx
@@ -13,6 +13,7 @@ import LazyImage from 'shared/LazyImage';
 import ListItem from 'shared/ListItem';
 import Svg from 'shared/Svg';
 import { dateToYear, formatDateRange } from 'utils/formatDate';
+import { formatAddress, formatGeolocationAddress } from 'utils/formatter';
 
 const TripsReceipt = ({ match }: RouterProps): JSX.Element => {
   return (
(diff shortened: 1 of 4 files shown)


tc:
can you uncommit your package-lock if no packages are added/removed?

jerminatorhits:
(out of scope of this PR) I noticed that these was a blank security deposit when booking:

image

 we should prob make provide a value of 0 or not render the security deposit at all if it doesn't exist

kcvan:
fixed the security deposit

src/components/routes/Trips/TripsReceipt/TripsReceipt.tsx
```diff
@@ -146,10 +146,11 @@ const TripsReceiptsMeta = (props: Booking): JSX.Element => {
                   <span>
                     {addressLine1 && `${addressLine1}, `}
```
tc: don't we have a address formatter for this?

kcvan: fixed! updated the formatter with the newer one from backend

PR #124, opened 2019-02-02
Jeremy/eng 957 mobile listing form

jerminatorhits:
## Description
Why did you write this code?
https://beetoken.atlassian.net/browse/ENG-957

## How to Test
- [ ] Visit https://dev02.beenest.io/host/listings/87134/listing_info
- [ ] Try filling out the form and navigating tabs on all screen sizes, esp mobile

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [x] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings
hyphens: auto is a thing. auto breaks words using a dictionary of sorts.
https://css-tricks.com/almanac/properties/h/hyphenate/

### DIFF
--- a/src/components/routes/Host/ListingForm/ListingForm.container.ts
+++ b/src/components/routes/Host/ListingForm/ListingForm.container.ts
@@ -113,7 +113,7 @@ const ListingFormDesktopContainer = styled(ListingFormTabletContainer)`
       width: 976px;
       min-width: 976px;
       form {
-        padding-left: 0;
+        padding: 40px 56px 96px 0;
       }
     }
   }
--- a/src/components/routes/Host/ListingForm/ListingForm.container.ts
+++ b/src/components/routes/Host/ListingForm/ListingForm.container.ts
@@ -6,13 +6,12 @@ const ListingFormMobileContainer = styled.section`
 
   & > .bee-general-wrapper {
     margin: 0 auto;
-    margin-bottom: 56px;
     min-height: calc(100% - 64px);
     width: 100%;
     min-width: 100%;
     form {
       min-height: calc(100% - 128px);
-      padding: 16px;
+      padding: 48px 24px 56px;
       width: 100%;
       label {
         margin-bottom: 8px;
--- a/src/components/routes/Host/ListingForm/ListingForm.container.ts
+++ b/src/components/routes/Host/ListingForm/ListingForm.container.ts
@@ -4,14 +4,14 @@ import { color } from 'styled/utils';
 const ListingFormMobileContainer = styled.section`
   width: 100%;
 
-
   & > .bee-general-wrapper {
-    min-height: calc(100% - 64px);
     margin: 0 auto;
+    margin-bottom: 56px;
+    min-height: calc(100% - 64px);
+    min-width: auto;
     form {
       min-height: calc(100% - 128px);
-      padding: 40px 0 96px;
-      padding-right: 56px;
+      padding: 16px;
       width: 100%;
       label {
         margin-bottom: 8px;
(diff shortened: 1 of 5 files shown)


src/components/routes/Host/ListingForm/ListingForm.container.ts
```diff
@@ -1,19 +1,18 @@
...
+const ListingFormMobileContainer = styled.section`
   width: 100%;
 
-
   & > .bee-general-wrapper {
-    min-height: calc(100% - 64px);
     margin: 0 auto;
+    margin-bottom: 56px;
+    min-height: calc(100% - 64px);
+    min-width: auto;
```
kcvan: i dont think `auto` is a valid value for min-width, i think you're looking for `0`, but i also think you don;'t need to add it if it's not being set anywhere else?

src/components/routes/Host/ListingForm/ListingForm.container.ts
```diff
@@ -96,4 +95,26 @@ const ListingFormContainer = styled.section`
...
+    }
+  }
+`;
+
+const ListingFormDesktopContainer = styled(ListingFormTabletContainer)`
+  @media (min-width: 1025px) {
+    & > .bee-general-wrapper {
+      width: 976px;
+      form {
+        padding-left: 0;
```
kcvan: for conformity, i would use `padding: 40px 56px 96px 0;`

src/components/routes/Host/ListingForm/ListingForm.tsx
```diff
@@ -192,7 +192,6 @@ class ListingForm extends React.Component<Props, State> {
                     onClick={() => {
                       if (!FormikProps.isValid) {
                         alert(`Cannot save changes due to errors:\n\n${Object.values(FormikProps.errors).join('\n').toString()}`);
```
kcvan: might need to change`\n` to `\n\r` fo windows

src/components/routes/Host/ListingForm/ListingFormNav/ListingFormNav.container.ts
```diff
@@ -1,72 +1,38 @@
...
+  width: 100%;
 
   .bee-general-wrapper {
     display: flex;
+    flex-direction: column;
     justify-content: space-between;
     height: 100%;
     margin: 0 auto;
-
+    min-width: auto;
```
kcvan: same here

src/components/routes/Host/ListingForm/ListingForm.container.ts
tc: if this is a consistent padding for all form containers, can it be a variable somewhere or just applied to `form` generally?

src/components/routes/Host/ListingForm/ListingForm.tsx
jerminatorhits: so with just using `\n` it looks like:

image

but with `\n\r` it looks like there are double spaces:

image

src/components/routes/Host/ListingForm/ListingFormNav/ListingFormNav.container.ts
jerminatorhits: removed width="100%" as props passed to GeneralWrapper and declared `width` and `min-width` here as well

kcvan:
running into stuff looking like this:
screen shot 2019-02-04 at 12 30 29 pm

1) the tabs look a little tall, maybe `64px`?
2) i think `Checkin` should be `Check-in`
3) we should allocate a space for the `Save & Exit` link so it's not on the same line as the first input titles
4) the form can use `padding: 48px 24px 56px;` (this might fix #3)
5) we can remove the `margin-bottom: 56px;` on the bee general wrapper because the new padding bottom from #4 will add this

src/components/routes/Host/ListingForm/ListingForm.container.ts
jerminatorhits: I think the `padding` was for the layout of this particular form -- though that would be something good to keep in mind a `<BeeFormContainer>` that has all the layouts/styles for a basic form.

jerminatorhits:
@kcvan could not replicate `1.` but made adjustments to accommodate the extra margin from `6.`
All other items should be updated

PR #129, opened 2019-02-04
kevin/avatar-hotfix

kcvan:
## Description
Why did you write this code?
fix(avatar): hotfix avatar


tc:
Hotfix since big avatar was blocking user

tc:
Should have proper fix in next pr

PR #131, opened 2019-02-04
Add Duplicate button

woeltjen:
## Description
Adds Duplicate as an option to listing cards in `/host/listings`, to make it easier to make more listings.

## How to Test
1. Navigate to `/host/listings` (and create at least one listing, if you don't have one)
2. Click "Duplicate" on a listing
3. Verify that a new listing named "Copy of..." appears
4. Verify that listing is initially unpublished

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
UX attention here is minimal; future enhancements (e.g. wait spinners) would make sense

### DIFF
--- a/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
+++ b/src/components/routes/Host/HostListingCard/HostListingCard.container.ts
@@ -35,7 +35,8 @@ const HostListingCardContainer = styled.article`
       width: 555px;
       a,
       button {
-        min-width: 125px;
+        min-width: 100px;
+        width: 100px;
       }
       label {
         display: inline-flex;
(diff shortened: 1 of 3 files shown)


woeltjen:
image

tc:
can we make all the buttons white bg with outline and only the Edit one colored?

woeltjen:
image

kcvan:
@tommy-bee i think the button color change is a little confusing.  the edit button is the same color as the other buttons when hovered, making it feel like the edit button is turned on

tc:
@kcvan we can remove the hover color if that helps. I just wanted to de-emphasize the least used buttons in favor of the most used one.

kcvan:
i think removing the hover might be a little bit of work, i didnt anticipate having a inverse button with no hover.  i think we can leave it for now and come back to it late then

PR #132, opened 2019-02-04
Remove ETH Payments contract references

woeltjen:
## Description
Following reimplementation of ETH payments via Unipay there is no separate ETH Payments contract, so remove references to its address & its ABI

## How to Test
Regression-test payment with ETH

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
May be a backend PR here, but after some cursory `grep`ping it looks like all the obsolete ETH payment flows have been reimplemented or refactored out already

## Learnings
Dijkstra's algorithm for shortest paths can be derived from a [lattice-linear predicate detection algorithm](https://arxiv.org/pdf/1812.10431.pdf) which is also useful as an archetype for certain O(n²) matching and pricing algorithms. (That has nothing to do with this PR, it's just the most interesting thing I've learned recently.)


kcvan:
did a quick find all for `ETH_PAYMENT_ADDRESS` to see if it's being used anywhere.  is it fine that we're still using it in the package.json?

woeltjen:
Good catch! I missed `package.json` there

PR #141, opened 2019-02-05
Paginate /admin/listings

woeltjen:
## Description
Allow paginating through listings in `/admin/listings` to help keep this page usable as our listing count grows

## How to Test
Test in concert with thebeetoken/beenest-backend#724

1. Go to `/admin/listings`
2. Verify that buttons to page through listings appear at bottom of page
3. Verify that paging forward through listings work
4. Verify that paging backward through listings works

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Pagination in bookings, perhaps?

## Learnings
There's an opportunity to refactor out our Tables into a shared component. We're only using tables in admin, though, so such a change would have limited impact

### DIFF
--- a/src/components/Admin/AdminListings/AdminListingsTable/AdminListingsTable.tsx
+++ b/src/components/Admin/AdminListings/AdminListingsTable/AdminListingsTable.tsx
@@ -56,7 +56,7 @@ const AdminListingsTable = ({ deleteListing }: Props): JSX.Element => {
                       </td>
                       <td>
                         <h6 className="admin-table-count">
-                          {offset + 1} to {offset + limit} of {count}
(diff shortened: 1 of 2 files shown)


woeltjen:
@kcvan Thanks! Fixed

image

kcvan:
also might be out of the scope of this pr but it would be nice for the screen to scroll to top after clicking the previous/next button instead of staying at the bottom of the page.  

everything else LGTM!

PR #146, opened 2019-02-06
Don't show "started" bookings under Trips

woeltjen:
## Description
Sending a user back to a `started` booking was implemented as a conversion-oriented convenience, but introduced some complexity: Allowing the user to complete that specific booking may violate business constraints (price quotes, reservations in particular) whereas sending the user to create a new booking leaves a `started` booking in our database and under the Trips tab.

Per standup, simplest solution is too remove `started` bookings from the Trips page.

## How to Test
1. Start to make a booking; do not pay!
2. Navigate to `/trips/current`
3. Verify that the booking you started does *not* appear

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
There is some related code that can be retired on the backend. We may also want to implement some different strategy to encourage completion of unfinished bookings.


kcvan:
can we add a default message if there are no current trips?

woeltjen:
> can we add a default message if there are no current trips?

That makes sense. Mind if I do that in a hotfix PR to keep the changesets narrow?

PR #147, opened 2019-02-06
show error when toggle publish fails

jerminatorhits:
## Description
Why did you write this code?
Goes with https://github.com/thebeetoken/beenest-backend/pull/732
Before, error is only thrown in the console. This pops up an alert when publish fails.

## How to Test
- [ ] Switch Backend branch to https://github.com/thebeetoken/beenest-backend/pull/732
- [ ] Signup using a fake email.
- [ ] Visit 'host/listings' and make a new listing
- [ ] Toggling the publish should fail

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/routes/Host/HostListingCard/HostListingCard.tsx
+++ b/src/components/routes/Host/HostListingCard/HostListingCard.tsx
@@ -79,7 +79,12 @@ const HostListingCard = (props: Props): JSX.Element => {
           <div className='host-listing-publish'>
             <label htmlFor={`publish-${id}`} title={canPublish ? '' : INCOMPLETE_LISTING}>
               <span className={canPublish ? '' : 'host-listing-meta--disabled'}>Publish</span>
-              <Switch checked={isActive} disabled={!canPublish} onColor={hexColor('correct')} onChange={() => toggleListing(id)} id={`publish-${id}`} />
+              <Switch
+                checked={isActive}
+                disabled={!canPublish}
+                onColor={hexColor('correct')}
+                onChange={() => toggleListing(id).catch((error) => alert(error))}
+                id={`publish-${id}`} />
             </label>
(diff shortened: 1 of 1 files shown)


jerminatorhits:
alert shows `You are not authorized to activate this listing` which could be misleading. Maybe there should be a prompt telling the user to verify and if so, should that text be placed in the front end or have it thrown via backend.

jerminatorhits:
@woeltjen makes sense to me. I'll include it in this PR!

One small possible confusion could be that the unverified user completed all the fields but still cannot publish, but maybe it's implied that they need to verify since the banner is there?
image

src/components/routes/Host/HostListingCard/HostListingCard.tsx
```diff
@@ -79,12 +80,18 @@ const HostListingCard = (props: Props): JSX.Element => {
...
+                      checked={isActive}
+                      disabled={!canPublish || !completedVerification}
+                      onColor={hexColor('correct')}
+                      onChange={() => toggleListing(id).catch((error) => alert(error))}
+                      id={`publish-${id}`} />
+                  );
+                }}
+              </FirebaseConsumer>
             </label>
             {!canPublish && <p className='host-listing-notice'>{INCOMPLETE_LISTING}</p>}
```
tc: maybe we can do ``!completedVerification ^^ "You must verify your email and phone to publish."``

jerminatorhits: 👍

PR #153, opened 2019-02-07
kevin/ENG-982

kcvan:
## Description
Why did you write this code?
fix(trips): fix trips display status and alter cryptoPortal to loadingPortal


jerminatorhits:
could you add `guest_rejected: 'Trip rejected by you'` in `ExpiredTripCard` in `cancelledDisplayMap` object? thx!

PR #155, opened 2019-02-08
kevin/ENG-961

kcvan:
## Description
Why did you write this code?
feat(google): add google auth and remove facebook auth except for login

to be used in conjunction with backend pr
## How to Test
- [ ] go to login/signup page
- [ ] click sign/login with google
- [ ] verify login works and everything is normal


woeltjen:
> * verify login works and everything is normal

One unexpected thing is that the avatar in the top-right got replaced with Google's default "V", but my own image still shows under profile:

image

I'm not sure that's worth trying to fix (or even calling a bug), but mentioning it for completeness.

kcvan:
@woeltjen if you refresh the page does it work? we've run into this issue where firebase doesnt update unless you refresh the page

woeltjen:
> @woeltjen if you refresh the page does it work? we've run into this issue where firebase doesnt update unless you refresh the page

Nope, the "V" stays a "V" after refresh.

kcvan:
@woeltjen thanks for the heads up, should work now

src/components/Header/Header.tsx
```diff
@@ -192,33 +194,69 @@ class DefaultMobileHeader extends React.Component<MobileHeaderProps> {
...
+                                if (screenType < ScreenType.TABLET) {
+                                  return (
+                                    <LazyImage
+                                      className={`${!showMenu ? 'show' : 'bee-lazy-image'}`}
+                                      height="24"
+                                      width="24"
+                                      src={
+                                        data.user.profilePicUrl
+                                          ? data.user.profilePicUrl
+                                          : user.photoURL
```
tc: don't understand this here

src/components/Header/Header.tsx
```diff
@@ -192,33 +194,69 @@ class DefaultMobileHeader extends React.Component<MobileHeaderProps> {
...
+                                    />
+                                  );
+                                }
+
+                                return (
+                                  <LazyImage
+                                    className={`${!showMenu ? 'show' : 'bee-lazy-image'}`}
+                                    height="48"
+                                    width="48"
+                                    src={
```
tc: here too, nested tenary is hard to read

src/components/Header/Header.tsx
kcvan: thanks, fixed!

src/components/Header/Header.tsx
kcvan: fixed!

PR #156, opened 2019-02-08
Jeremy/eng 984

jerminatorhits:
## Description
Why did you write this code?
Trip cards should show Lat Lng when street address is not available. This matches the Receipt Page behavior. (If the address gets really long we might need to go to 2 lines)

image

## How to Test
- [ ] Book a Thailand booking
- [ ] Visit /trips and see lat lng

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/routes/Trips/ActiveTripCard.tsx
+++ b/src/components/routes/Trips/ActiveTripCard.tsx
@@ -15,7 +15,7 @@ import { getUserBookingDisplayStatus } from 'utils/bookingsDisplayStatus';
 import { ToggleProvider, ToggleProviderRef } from 'shared/ToggleProvider';
 import Portal from 'shared/Portal';
 import ContactHostForm from 'shared/ContactHostForm';
-import { getGoogleMapURI, formatAddress } from 'utils/formatter';
+import { getGoogleMapURI, formatAddress, formatGeolocationAddress } from 'utils/formatter';
 
 interface Props {
   onCancelClick: () => void;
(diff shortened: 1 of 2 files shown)


tc:
can you have it just as ``Lat, Lng: X, Y`` rather than ``Lat: X, Lng: Y`` so it can be copy-pasta easily?

jerminatorhits:
changed to `Lat, Lng: ${lat}, ${lng}`

jerminatorhits:
actually it might be harder to read:
image

I could also put it in a separate line

src/components/routes/Trips/ActiveTripCard.tsx
```diff
@@ -40,7 +41,7 @@ const ActiveTripCard = ({ onCancelClick, trip }: Props) => {
         <div className="address">
         <BeeLink href={getGoogleMapURI(listing)} target="_blank">
           <ListItem noHover suffixColor="secondary" textColor="secondary" textTransform="uppercase">
-            <span>{formatAddress(streetAddress, listing.city, listing.state, listing.country.toUpperCase())}</span>
+            <span>{formatAddress(streetAddress || `Lat, Lng: ${lat}, ${lng}`, listing.city, listing.state, listing.country.toUpperCase())}</span>
```
kcvan: i think theres a `formatGeolocationAddress` you can use

jerminatorhits: 🙏

PR #161, opened 2019-02-09
Jeremy/eng 701

jerminatorhits:
## Description
Why did you write this code?
https://beetoken.atlassian.net/browse/ENG-701
fetchCoordinates tries to setState when GoogleMaps is already unmounted (user has navigated to another tab)

Small Intro to Hooks: https://reactjs.org/docs/hooks-overview.html#%EF%B8%8F%EF%B8%8F-effect-hook-%EF%B8%8F-effect-hook

Still ended up using the `isMounted` flag since there wasn't an easy way to "unsubscribe" from the promise.

## How to Test
- [ ] Visit /host/listings/:id/edit
- [ ] Click from 'listing_info' to other tabs
- [ ] Verify error in console no longer shows up

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

## Further Work
Will this have future work?
This is my first stab at hooks. Probably more maintenance required if Hooks become more commonplace.

## Learnings
NullIsland is a thing `{ lat: 0, lng: 0 }`, used it to satisfy typescript demands: https://en.wikipedia.org/wiki/Null_Island


src/components/shared/GoogleMaps/GoogleMaps.tsx
```diff
@@ -40,71 +29,65 @@ const circleOptions = {
...
+
+        console.error(error);
+        setLoading(false);
+        setError(true);
+      });
+      
+    return cleanUp;
+  },
+    [debouncedAddress]
+  );
```
woeltjen: Indentation here got a little confusing for me; think this might be clearer as:

    }, [debouncedAddress]);

...so that it goes back to (and closes) the indentation level of `useEffect` on a single line

jerminatorhits: prettifier agrees! good catch

woeltjen:
Noticed a regression in `/listings/:id`

image

Looks this way for any listing I view on this branch

jerminatorhits:
yikes! thanks @woeltjen will put in a fix!

tc:
conflicts

PR #169, opened 2019-02-11
kevin/work-theme

kcvan:
## Description
Why did you write this code?
feat(theme): add theme
## How to Test
- [ ] Visit /work
- [ ] Verify that there are styles on the page

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android


kcvan:
should look like this in dev console:
screen shot 2019-02-11 at 12 26 54 pm

src/work.tsx
```diff
@@ -10,6 +10,7 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { BrowserRouter } from 'react-router-dom';
 import 'bootstrap/dist/css/bootstrap.css';
+import 'styled/theme.css';
```
woeltjen: I'm a bit confused. `theme.css` looks like Boostrap; why are we checking it in here instead of reading it from `node_modules`?

kcvan: this is the `front` theme overriding some of the default bootstrap themes

tc: if you look at the docs front-v2.2/documentation/index.html#starter-template
they only include the theme and not the bootstrap.css file

tc:
I was thinking of a different directory structure for rebrand:
```
src
src/components-work (temporary, becomes src/components)
src/pages
src/pages/account
src/pages/booking
src/pages/trips
```

``src/pages`` replaces ``src/components/routes`` and represents actual pages.
``src/pages`` will be lowercase so it resembles the path url.

cc @woeltjen @kcvan @jerminatorhits

src/work.tsx
kcvan: did a search for styles that bootstrap has that the template doesnt, looks like all the styled from bootstrap are in the template + more! removing the bootstrap.css from the file

kcvan: actually, it looked like the new 4.3 bootstrap came out with 1000 lines of new css. should we still keep the bootstrap.css in case we want to use some of their new class names?

tc: i would keep just `theme.css` to avoid any conflicts. the file specifies the bootstrap version at the top

kcvan:
@tommy-bee so `src/routes` will be replaced with `src/pages`?  I was wondering if this would adhere to react file naming where there component is usually capitalized.  So instead of `src/pages/account` we would have `src/pages/Account`.  But going this route looks like we would just be renaming `routes` to `pages`, so not sure what to do here

src/work.tsx
kcvan: 👍 we can cherry pick any styles we want from the new version and just add it to the `theme.css` file

tc: why not pull in the scss file instead of the css file so we can use set the color/font variables?

tc:
for your example, it would be something like: ``src/pages/account/account``
i'm suggesting this to: 
a. make the directory more similar to url routes
b. distinguish high level pages and the components (living in the components dir) that compose them

tc:
@kcvan can we just remove the bootstrap npm package if we don't use it?

kcvan:
@tommy-bee it looks like the `front` scss file still pulls from `bootstrap` core files

kcvan:
i think the `src/pages/account/account` change is fine.  i think they use that path structure because theyre mainly used for ssr so maybe it's to imitate backend routing?

src/work.tsx
kcvan: good call! using scss files

woeltjen:
From [the license](https://themes.getbootstrap.com/licenses/#fullExtendedLicense) (assuming we're talking about the same [front](https://themes.getbootstrap.com/product/front-multipurpose-responsive-template/)):

> *You cannot…*
> ...
> Allow the End User of the End Product to access the Theme’s source and use it separately from the End Product.

Checking the theme into our open source repo looks like a license violation to me. (Not sure if we got the Standard or Extended License, but the same clause appears in both.)

We could check it into a private repo (e.g. `beenest-theme`) and add it as a dependency that way. It will make builds painful in some circumstances, so I'm open to better ideas.

kcvan:
closing out PR for further discussion

PR #177, opened 2019-02-12
Show message for two-day crypto bookings.

woeltjen:
## Description
Hiding BEE/ETH/DAI then a booking is less than two days in advance could be confusing for users. 

Instead, show BEE/ETH/DAI as options, but show a message and disable the button to continue.

## How to Test
1. Start to make a booking less than two days in advance
    1. Note that host must have an eth wallet
2. Verify that BEE, ETH, and DAI all appear as options
3. Select those options
4. Verify that button to continue booking is disabled, and that an informative message appears
5. Start a booking more than two days in advance
6. Verify that BEE/ETH/DAI can be paid for such a booking

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/components/routes/Booking/BookingOptions/BookingOptionsCrypto/BookingOptionsCrypto.tsx
+++ b/src/components/routes/Booking/BookingOptions/BookingOptionsCrypto/BookingOptionsCrypto.tsx
@@ -110,7 +110,7 @@ const BookingOptionsCrypto = ({ booking, currency, fromBee, history }: Props) =>
                     </div>
                   ))}
                 {!isTwoDaysFromNow && <div className="booking-options-error">
-                  <p>Bookings paid using Ethereum or ERC-20 tokens must be made at least two days in advance.</p>
+                  <p>Bookings paid using Ethereum or ERC-20 tokens must be made more than two days in advance.</p>
                 </div>}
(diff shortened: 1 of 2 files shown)


src/components/routes/Booking/BookingOptions/BookingOptionsCrypto/BookingOptionsCrypto.tsx
```diff
@@ -104,6 +109,9 @@ const BookingOptionsCrypto = ({ booking, currency, fromBee, history }: Props) =>
                       </div>
                     </div>
                   ))}
+                {!isTwoDaysFromNow && <div className="booking-options-error">
+                  <p>Bookings paid using Ethereum or ERC-20 tokens must be made at least two days in advance.</p>
```
jerminatorhits: Today (2/12), I chose dates 2/14 to 2/15 and the message shows up.

If it says at least 2 days in advance, one might think those dates are technically 2 days in advance. Maybe it can be "more than two days in advance"?

PR #178, opened 2019-02-12
kevin/theme-cdn

kcvan:
## Description
Why did you write this code?
feat(cdn,bootstrap): add css file by cdn and remove bootstrap
## How to Test
- [ ] Visit /work
- [ ] Verify button color is `#00c9a7`

### DIFF
--- a/src/templates/work.development.html
+++ b/src/templates/work.development.html
@@ -3,10 +3,10 @@
 
 <head>
   <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="Beenest" content="Beenest Platform">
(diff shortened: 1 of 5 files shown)


src/templates/work.development.html
```diff
@@ -3,10 +3,10 @@
 
 <head>
   <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="Beenest" content="Beenest Platform">
   <title>Beenest</title>
-  <link href="https://fonts.googleapis.com/css?family=Roboto:100,400,500,700,900" rel="stylesheet">
```
tc: do we no longer need the font? or is it embedded in the css file/

kcvan: embedded

PR #180, opened 2019-02-12
add fetch policy to get most up-to-date info from backend

jerminatorhits:
## Description
Why did you write this code?
When a user submits a booking request, then navigates to the listing's page, the dates are not blocked out. User has to refresh to see the updated blockout dates.

## How to Test
- [ ] Make a booking request
- [ ] Navigate to listing page to see those selected dates blocked.

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/routes/Listing/Listing/Listing.tsx
+++ b/src/components/routes/Listing/Listing/Listing.tsx
@@ -28,7 +28,7 @@ class Listing extends React.Component<RouterProps> {
     const { id } = this.props.match.params;
     return (
       <ListingContainer>
-        <Query query={GET_PUBLIC_LISTING} variables={{ id, ...PROFILE_IMAGE_PARAMETERS }}>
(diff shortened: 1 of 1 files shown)


tc:
would `cache-and-network` still grab data from the cache?

jerminatorhits:
hmm technically yes, but the Apollo docs aren't really clear if after the network fetch, they update the cache and if that data updates the UI. https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-config-options-fetchPolicy. From my understanding, it uses the cache at first while waiting for the network fetch data and then updates everything once received.

The calendar does update with the new dates using `cache-and-network` so it works via smoke testing. If we want to be safe, we can use `network-only`.

PR #181, opened 2019-02-12
Jeremy/eng 989 - Listing Form updates

jerminatorhits:
## Description
Why did you write this code?
update listing help section
- include 1 good and 1 bad photo instead of 2 good ones.
- more listing title examples
- change Button into Audio Loading so navigating between forms is more fluid

- Lmk if there are any suggestion on more things to add/change!

## How to Test
- [ ] Visit host/listings/:id/edit OR https://dev03.beenest.io/host/listings/256/listing_info
- [ ] See help section for `listing title` and `cover photo`
- [ ] Ensure it displays well with various screen sizes.

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?
Add dynamic disclaimer at bottom to let ppl know if the listing is ready to publish. This is done on the backend to make sure fields exist and is attached to the `canPublish` attribute. However, the front end would need a real-time evaluator based on the form field values to execute this feature well. This function will need to match the backend criteria and add another second place to maintain what publish validity means. Will decide if this is worth doing and possibly submit it in a future PR

## Learnings

### DIFF
--- a/src/components/routes/Host/ListingForm/ListingHelp.tsx
+++ b/src/components/routes/Host/ListingForm/ListingHelp.tsx
@@ -23,10 +23,12 @@ const HomeType = () => (
 const Title = () => (
   <div>
     <h3>Listing Name</h3>
-    <h3>Choose a name that perfectly describes your listing. Be sure to use proper casing.</h3>
-    <p>&quot;Beautiful Home with View of Manhattan Skyline&quot;</p>
-    <p>&quot;Convenient Room in the Heart of San Francisco&quot;</p>
-    <p>&quot;Downtown LA Apartment near Pershing Square&quot;</p>
+    <p>Choose a name that perfectly describes your listing. Be sure to use proper casing.</p>
+    <ul>
+      <li>&quot;Beautiful Home with View of Manhattan Skyline&quot;</li>
+      <li>&quot;Convenient Room in the Heart of San Francisco&quot;</li>
+      <li>&quot;Downtown LA Apartment near Pershing Square&quot;</li>
+    </ul>
   </div>
 );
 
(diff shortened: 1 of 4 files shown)


src/components/routes/Host/ListingForm/ListingHelp.tsx
```diff
@@ -23,7 +23,10 @@ const HomeType = () => (
 const Title = () => (
   <div>
     <h3>Listing Name</h3>
-    <p>&quot;Beautiful Home with Wifi and Views&quot;</p>
+    <h3>Choose a name that perfectly describes your listing. Be sure to use proper casing.</h3>
+    <p>&quot;Beautiful Home with View of Manhattan Skyline&quot;</p>
+    <p>&quot;Convenient Room in the Heart of San Francisco&quot;</p>
+    <p>&quot;Downtown LA Apartment near Pershing Square&quot;</p>
```
woeltjen: The fontography here gets a little funny with the two `h3`s

What about `<h3>` for Listing Name, `<p>` for "Choose a name...", then maybe a `<ul>` for the examples? That seems consistent with the other help text

jerminatorhits:
lol sry. will make those changes quick!

src/components/routes/Host/ListingForm/ListingHelp.tsx
jerminatorhits: good idea. thanks!

PR #184, opened 2019-02-13
Account Page WIP

jerminatorhits:
## Description
Why did you write this code?

## How to Test
- [ ] Visit XXX
- [ ] Click Y

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


src/pages/index.tsx
```diff
@@ -2,10 +2,12 @@ import * as React from 'react';
 import { Route, Switch } from 'react-router-dom';
 
 import Home from './home';
+import WorkAccount from './WorkAccount';
```
tc: you can just call this `Account`; it's in a separate dir from the other components and it'll make it easier once it port it.

jerminatorhits: 👍

src/pages/Account/AccountGeneral.tsx
```diff
@@ -0,0 +1,148 @@
...
+                  placeholder="First name"
+                  type="text" />
+                <FormFeedback>{errors.firstName}</FormFeedback>
+              </FormGroup>
+            </Col>
+            <Col md={6}>
+              <FormGroup inline>
+                <Label for={UserField.LAST_NAME} className="form-label">Last Name</Label>
+                <Field
+                  className={`form-control${errors.lastName ? ' is-invalid' : ''}`}
```
kcvan: we can use reactstrap `Input` and get pretty colors and stylings

src/pages/Account/Account.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    <NavLink href="/work/account/verification">Verification</NavLink>
+                  </NavItem>
+                </Nav>
+                
+                <Container>
+                  <Row>
+                    <Col md={6}>
+                      <hr />
+                      <Switch>
+                        <Route exact path="/work/account/general" render={(props: RouterProps) => <AccountGeneral {...props} user={user} />} />
```
tc: there's also https://github.com/reactjs/react-tabs for tabs

src/pages/Account/AccountGeneral.tsx
jerminatorhits: swapped

src/pages/Account/Account.tsx
jerminatorhits: was able to 'reactstraps' tabs in an effort to maintain consistent styling.

src/pages/Account/AccountGeneral.tsx
```diff
@@ -0,0 +1,167 @@
...
+  [UserField.FIRST_NAME]: Yup.string()
+    .required('Please provide your first name.'),
+  [UserField.LAST_NAME]: Yup.string()
+    .required('Please provide your last name.'),
+  [UserField.EMAIL]: Yup.string(),
+  [UserField.ABOUT]: Yup.string()
+    .required('Please fill out the About section.'),
+})
+
+const defaultValues: FormValues = {
```
woeltjen: Would it be a useful simplification to do something like this?

    const defaultValues = Object.key(UserField).reduce(
      (values, key) => ({ ...values, [UserField[key]]: '' })
    );

That would reduce the number of places we'd need to update when adding a new form field, tho it only makes sense to do that if defaults will always be empty strings

jerminatorhits: good thing to consider, though I feel each field value declared explicitly feels a bit easier to read.

If other forms do include numbers/arrays/booleans, it would be easier to add / adjust at first glance.

jerminatorhits:
closing to split into smaller PRs

PR #189, opened 2019-02-16
kevin/ENG-993

kcvan:
## Description
Why did you write this code?
feat(signup): add signup page
## How to Test
- [ ] Visit /work/signup
- [ ] check that errors pop up accordingly
- [ ] sign in with google should work
- [ ] signing in should redirect to /work

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android


src/pages/signup/Signup.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { FirebaseConsumer, FirebaseUserProps } from 'HOCs/FirebaseProvider';
+
+import SignupForm from './SignupForm';
+
+const Signup = (props: RouteProps) => (
+  <FirebaseConsumer>
+    {({ loading, user }: FirebaseUserProps) => {
+      if (loading) {
+        return (
+          <Container className="d-flex align-items-center justify-content-center position-relative height-100vh">
```
kcvan: need to use styled components to make reusable containers.  will do this in separate pr

src/pages/signup/Signup.tsx
```diff
@@ -0,0 +1,53 @@
...
+        );
+      }
+
+      if (user) {
+        const state = (props.location && props.location.state) || {};
+        const destination = state.referrer || '/work';
+        return <Redirect to={destination} />;
+      }
+      return (
+        <Container className="d-flex flex-column flex-lg-row align-items-center position-relative height-lg-100vh px-0" fluid>
```
kcvan: here too

src/pages/signup/SignupForm.tsx
```diff
@@ -0,0 +1,270 @@
...
+        {({ errors, touched, setFieldValue, submitForm, isSubmitting }) => (
+          <Form method="POST">
+            <div className="mb-7">
+              <h2 className="h3 text-primary font-weight-normal mb-0">
+                Welcome to <span className="font-weight-semi-bold">Beenest</span>
+              </h2>
+              <p>Fill out the form to get started.</p>
+            </div>
+
+            <Row className="d-flex flex-sm-column flex-md-row">
```
kcvan: i was thinking about moving these details to a helper array of objects and map it out to avoid boilerplate bloat

src/pages/signup/Signup.tsx
kcvan: addressed as string variable

src/pages/signup/Signup.tsx
kcvan: addressed

src/pages/signup/SignupForm.tsx
kcvan: actually might be too many variables to cover to cleanly do this at this moment

src/components-work/AuthenticationTestimonials/AuthenticationTestimonials.tsx
```diff
@@ -18,7 +18,7 @@ const items = [
   },
 ];
 
-class LoginTestimonials extends React.Component {
+class AuthenticationTestimonials extends React.Component {
```
tc: we should just make this `GuestTestimonials` or `HostTestimonials`; AuthTestimonials doesn't make sense?

src/components/work/AuthenticationTestimonials/AuthenticationTestimonials.tsx
```diff
@@ -0,0 +1,78 @@
...
+    altText: 'This could be the second testimonial',
+    caption: 'Slide 2',
+  },
+  {
+    altText: 'Sometimes this is the third testimonial',
+    caption: 'Slide 3',
+  },
+];
+
+class AuthenticationTestimonials extends React.Component {
```
tc: can we shorten the prefix to `Auth` instead of `Authentication`? It has the same meaning and a lot shorter to type.

src/components-work/AuthenticationTestimonials/AuthenticationTestimonials.tsx
kcvan: wasnt sure if we were actually going to use this testimonial carousel, wiull rename to `UserTestimonials`

src/components/work/AuthenticationTestimonials/AuthenticationTestimonials.tsx
kcvan: 👍

src/pages/signup/SignupForm.tsx
```diff
@@ -0,0 +1,270 @@
...
+class SignupForm extends React.Component<SignupProps, State> {
+  readonly state = {
+    providerError: null,
+  };
+
+  render() {
+    return (
+      <Formik
+        initialValues={{
+          signupEmail: '',
```
tc: why do we need the signup prefix?

src/pages/signup/SignupForm.tsx
```diff
@@ -0,0 +1,270 @@
...
+                  </Label>
+                  <Input
+                    type="text"
+                    name="signupFirstName"
+                    id="signupFirstName"
+                    tag={Field}
+                    onChange={(event: React.FormEvent<HTMLInputElement>) => {
+                      setFieldValue('signupFirstName', event.currentTarget.value);
+
+                      if (this.state.providerError !== null) {
```
tc: why is the conditional needed?

src/pages/signup/SignupForm.tsx
```diff
@@ -0,0 +1,270 @@
...
+                }}
+                placeholder="********"
+                invalid={!!errors.signupPassword && touched.signupPassword}
+              />
+              <FormFeedback>{errors.signupPassword}</FormFeedback>
+            </FormGroup>
+
+            <FormGroup>
+              <Label for="signupConfirmPassword" className="form-label">
+                Confirm Password
```
tc: do we need a confirm password?

src/pages/signup/SignupForm.tsx
kcvan: i wanted to has a prefix because ids should be unique throughout the app, so having something like `firstName` would probably show up in more than one form on the app

tc: why would this need to be unique throughout the app?

src/pages/signup/SignupForm.tsx
kcvan: added the conditional so it wouldnt set state if it was already null to prevent extra set states from leaking

src/pages/signup/SignupForm.tsx
kcvan: following convention of classes = reusable and id = unique

https://developer.mozilla.org/en-US/docs/Web/API/Element/id

src/pages/signup/SignupForm.tsx
kcvan: i dont think it's bad to have it.  i think users expect/dont mind a confirm password field, i feel it adds a more secure feeling to the site.  we can remove if the design calls for it not being there?

src/pages/signup/SignupForm.tsx
kcvan: added the conditional because formik/reactstrap isnt good at handling server errors so we us our own state (vs formiks) to show an error from our server/firebase.  formik will reset the error fields on it's own but since formik has it's own state and validators it isnt hooked up to our server/firebase error state

src/pages/signup/Signup.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import SignupForm from './SignupForm';
+
+const Signup = (props: RouteProps) => (
+  <FirebaseConsumer>
+    {({ loading, user }: FirebaseUserProps) => {
+      if (loading) {
+        return (
+          <Container className={VIEWPORT_CENTER_LAYOUT}>
+            <Spinner color="primary" style={{ width: '8rem', height: '8rem' }} type="grow" />
```
tc: can you hide this behind a "Loading" component so it's consistent

src/pages/signup/SignupForm.tsx
tc: "it must be unique in a document"
I think it's saying the id should be unique to a page not the entire app. if we have to prefix every id, our ids are going to be reallyLong.

src/pages/signup/SignupForm.tsx
tc: can this be a generic function so each of the inputs can re-use the function?
```
handleChange:
   setFieldValue(htmlId, event.currentTarget.value); 
...
```

src/pages/signup/SignupForm.tsx
kcvan: i think it's supposed to be the entire app? getElementById only returns one element that matches the id, where as getElementsByClass returns an array of elements

src/pages/signup/SignupForm.tsx
kcvan: 👍

src/pages/signup/Signup.tsx
kcvan: 👍

src/pages/login/LoginForm.tsx
```diff
@@ -154,7 +155,7 @@ class LoginForm extends React.Component<LoginProps, State> {
 
   handleSubmit = (values: LoginFormInput, actions: FormikActions<LoginFormInput>) => {
     login(values.loginEmail, values.loginPassword).catch(error => {
-      actions.setErrors({ authenticationError: error.message });
+      actions.setErrors({ authenticationError: getFriendlyErrorMessage(error) });
```
tc: instead of authenticationError, you can name this `submitError` and it can be re-used as the attribute name for submission errors.

src/pages/login/Login.tsx
```diff
@@ -2,18 +2,20 @@ import * as React from 'react';
...
 import LoginForm from './LoginForm';
-import LoginTestimonals from './LoginTestimonials';
 
 const Login = (props: RouteProps) => (
   <FirebaseConsumer>
     {({ loading, user }: FirebaseUserProps) => {
       if (loading) {
         return (
-          <Container className="d-flex align-items-center justify-content-center position-relative height-100vh">
+          <Container className={VIEWPORT_CENTER_LAYOUT}>
```
tc: The other way of doing this is `<CenterLayoutContainer>` instead of the css classes in a string so the programmer doesn't have to think whether to use a "Container" or a "Row" or a "div".

Unless the intended behavior is that the classes should be applied independent of the component

src/pages/login/Login.tsx
```diff
@@ -25,9 +27,9 @@ const Login = (props: RouteProps) => (
         return <Redirect to={destination} />;
       }
       return (
-        <Container className="d-flex align-items-center position-relative height-lg-100vh px-0" fluid>
-          <AuthenticationHeader />
-          <LoginTestimonals />
+        <Container className={AUTHENTICATION_CONTAINER} fluid>
```
tc: Maybe replace all ``AUTHENTICATION`` with `AUTH`?

src/pages/login/LoginForm.tsx
kcvan: 👍

src/pages/login/Login.tsx
kcvan: 👍

src/pages/signup/SignupForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+      .then((_) => login(input.email, input.password))
+      .then((_) => !auth.currentUser ? Promise.resolve() : auth.currentUser.sendEmailVerification())
+      .catch((error: Error) => {
+        actions.setErrors({ [SUBMIT_ERROR]: getFriendlyErrorMessage(error) });
+        actions.setSubmitting(false);
+      });
+  };
+
+  handleChange = (htmlId: string, event: React.FormEvent<HTMLInputElement>, setFieldValue: (field: string, value: any) => void) => {
+    setFieldValue(htmlId, event.currentTarget.value);
```
tc: can the html id come from the event.currentTarget?

src/pages/login/LoginForm.tsx
```diff
@@ -36,48 +47,47 @@ class LoginForm extends React.Component<LoginProps, State> {
...
+          [EMAIL]: '',
+          [PASSWORD]: '',
+          [SUBMIT_ERROR]: '',
         }}
         validationSchema={LoginSchema}
         onSubmit={this.handleSubmit}
       >
         {({ errors, touched, setFieldValue, submitForm, isSubmitting }) => (
-          <Form className="mt-5" method="POST">
+          <Form method="POST">
```
tc: is the method="POST" necessary?

kcvan: yeah if we don't have method="POST" it assumes get request and will append the input into into the url parameters

tc: 👍

src/pages/signup/SignupForm.tsx
kcvan: nice was able to get it from the `currentTarget.name`~

src/pages/login/LoginForm.tsx
kcvan: reactstrap's `Form` component is pretty basic so it needs things like this

tc: Why not use Formik's Form component?

kcvan: actually reactstrap has a nifty `tag` prop that takes an a component.  was able to alias formik's `Form` as `FormikForm` and added it in as the tag prop

kcvan: also we're using as much reactstrap as possible so we get the right stylings right off the bat

src/pages/signup/Signup.tsx
kcvan: i'll address this in a separate PR

PR #191, opened 2019-02-18
kevin/work-test

kcvan:
## Description
Why did you write this code?
fix work route


tsconfig.json
```diff
@@ -28,6 +28,7 @@
       "ABIs/*": ["ABIs/*"],
       "assets/*": ["assets/*"],
       "components/*": ["components/*"],
+      "componentsWork/*": ["componentsWork/*"],
```
tc: would it be easier to just put the rebrand components into `/components/work` and then just remove the `work` directory once they replace the other ones?

src/pages/Work.tsx
```diff
@@ -11,6 +11,7 @@ const Work = () => (
       <Route path="/work/account" component={Account} />
       <Route exact path="/work/login" component={Login} />
       <Route path="/work" component={Home} />
+      <Redirect from="/" to="/work" />
```
tc: you shouldn't need to do this?

PR #193, opened 2019-02-18
Jeremy/eng 998 payment

jerminatorhits:
## Description
Why did you write this code?
Implement Account Payments tab (add / delete) credit cards

## How to Test
- [ ] Visit /work/account/payment
- [ ] Add a card successfully
- [ ] Delete a card successfully
- [ ] Verify card can be used to book

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


src/pages/Account/NewCardForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+      [NewCardField.CARD_CVC]: false,
+      [NewCardField.NAME]: false,
+      [NewCardField.POSTAL_CODE]: false,
+    },
+    errors: {
+      [NewCardField.CARD_EXPIRY]: '',
+      [NewCardField.CARD_NUMBER]: '',
+      [NewCardField.CARD_CVC]: '',
+      [NewCardField.NAME]: '',
+      [NewCardField.POSTAL_CODE]: '',
```
woeltjen: Think that `completed` and `errors` can be derived from `Object.keys(NewCardField)` to reduce duplication here

src/pages/Account/NewCardForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+    [key: string]: boolean;
+  },
+  errors: {
+    cardExpiry: string;
+    cardNumber: string;
+    cardCvc: string;
+    name: string;
+    postalCode: string;
+    submit: string;
+    [key: string]: string;
```
woeltjen: For `completed` and `errors`, do we lose anything useful if we just say `[key: string]: boolean` and `[key: string]: string` and omit the fields? (Even better, can these be `[key: NewCardField]`? I feel like we found out the answer was *no* last time I wondered that...)

More broadly, from here to where state is initialized we list out our form properties five times, and that's all just boilerplate. Would be great to figure out how to get that down to a single schema/interface/enum/something here, with one of our smaller forms, and hopefully come out with an approach we can use elsewhere

src/pages/Account/DeleteCardForm.tsx
```diff
@@ -0,0 +1,66 @@
...
+    return deletePaymentSource(paymentSource.id)
+      .then(() => {
+        handleModal();
+        setAlert(`Success! Your card ${paymentSource.stripeBrand} ending in ${paymentSource.stripeLast4} has been deleted.`);
+      })
+      .catch((error: any) => alert('Oops! Something went wrong:\n\n' + error));
+  }
+  return (
+    <>
+      <ModalHeader>Delete Card</ModalHeader>
```
woeltjen: It's a minor point, but for `NewCardForm` the `ModalHeader` and `ModalBody` are added by `AccountPayment`, whereas here they're part of the form component. For consistency, I'd use the approach from `NewCardForm` in both places. (It's also a bit odd that `ModalHeader` and so on are in here without a `Modal`; it's a bit non-intuitive that `DeleteCardForm` depends on providing a wrapper `Modal` externally.)

src/pages/Account/NewCardForm.tsx
jerminatorhits: thanks, updated!

src/pages/Account/NewCardForm.tsx
jerminatorhits: removed the explicit type declarations.
> Even better, can these be `[key: NewCardField]`?
ya typescript doesn't like that =(

Would love to reduce the boilerplate! Will keep thinking of better solutions and am open to suggestions. This form may be a special case since Stripe Elements abstract a lot of the implementation and Formik isn't used.

src/pages/Account/DeleteCardForm.tsx
jerminatorhits: Agreed - consistency should be maintained.

One thing that made me feel uneasy was the `<ModalFooter />` component that goes after the ModalBody has some buttons with networking logic in it. (e.g. handleDelete). The `<NewCardForm />` doesn't even have a `<ModalFooter />` which might go against the framework's intentions.

I'm thinking to just make the <Modal><Form /></Modal> into one component. If the Form needs to be extracted somewhere down the line, it shouldn't be too difficult to pull out.

src/pages/Account/AccountPayment.tsx
```diff
@@ -1,10 +1,91 @@
 import * as React from 'react';
+import { ListGroup, ListGroupItem, Row, Col, Modal, ModalHeader, ModalBody, Alert } from 'reactstrap';
+import { Query } from 'react-apollo';
+import { GET_PAYMENT_SOURCES, PaymentSource } from 'networking/paymentSources';
+import AudioLoading from 'shared/loading/AudioLoading';
+import { CreditBalance } from 'networking/users';
+import NewCardForm from './NewCardForm';
```
kcvan: gimme dat spacing

src/pages/Account/DeleteCardForm.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const DeleteCardForm = ({ deletePaymentSource, handleModal, paymentSource, setAlert }: Props) => {
+  const handleDelete = () => {
+    return deletePaymentSource(paymentSource.id)
+      .then(() => {
+        handleModal();
+        setAlert(`Success! Your card ${paymentSource.stripeBrand} ending in ${paymentSource.stripeLast4} has been deleted.`);
+      })
+      .catch((error: any) => alert('Oops! Something went wrong:\n\n' + error));
```
kcvan: i think we can use the `Alert` component from reactstrap now

src/pages/Account/AccountPayment.tsx
jerminatorhits: 👍

src/pages/Account/DeleteCardForm.tsx
jerminatorhits: 👍

src/components/work/Alert/Alert.tsx
```diff
@@ -0,0 +1,14 @@
...
+  msg: string,
+  color: string,
+  show: boolean,
+}
+
+const Alert = (props: AlertProperties) => (
+  <ReactStrapAlert {...props} />
+);
+
+export default Alert;
```
jerminatorhits: this was mainly created for a central place to declare the AlertProperties interface used in several places.

src/pages/Account/DeleteCardForm.tsx
jerminatorhits: Insightful finding:

After doing some research on Modal implementation on ReactStrap, it looks like the convention is that Modals are supposed to be rendered at all times but use the `isOpen` prop to control them. (It may not matter too much for modals, but for `Alert`, I had trouble with the fade out animation because I was conditionally rendering the whole component).

src/pages/Account/AccountPayment.tsx
```diff
@@ -1,10 +1,96 @@
...
+              </Col>
+            </Row>
+            <h4>Credit Cards:</h4>
+            <ListGroup className="mb-2 d-flex flex-column">{renderPaymentSources}</ListGroup>
+
+            <Row>
+              <Col xs="12">
+                <div onClick={() => handleModal(ModalType.ADD_NEW_CARD)} className="w-auto d-inline-block align-items-center">
+                  <i className="fas fa-plus-circle" />
+                  <h6 className="ml-2 mb-0 d-inline-block">Add New Card</h6>
```
tc: links should be `a` tag

src/pages/Account/NewCardForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+          },
+        });
+      },
+    }),
+  })
+)(injectStripe(NewCardForm));
+
+const StripeWrappedNewCardForm = ({ handleModal, setAlert }: any) => (
+  <StripeWrapper>
+    <EnhancedNewCardForm handleModal={handleModal} setAlert={setAlert} />
```
tc: can you clarify what `handleModal` does? is handleModalOpen? or doing something else

jerminatorhits: `handleModal` sets the modal for the page. It takes 2 params: ModalType enum and a sets the current paymentSource if necessary (for the DeleteCard). It can control which type of modal to open as well as close modals if no arg is given.

```
function handleModal(modal?: ModalType, newPaymentSource?: PaymentSource) {
    setModal(modal)
    newPaymentSource && setPaymentSource(newPaymentSource); // null check prevents modal from rendering undefined paymentSource when fading out
  }
```

src/pages/Account/AccountPayment.tsx
jerminatorhits: there is no link present

tc: shouldn't L64 be an `a` tag vs `div` tag? or a `a` tag inside the `div`?
It is a clickable element (which is a link)

src/pages/Account/NewCardForm.tsx
tc: maybe `handleModalAction` ?

src/pages/Account/AccountPayment.tsx
kcvan: are we trying to make just the text clickable or the entire div? if its just the text then i think `a` tag is good, if it;s the entire div we can just wrap the div in a `a` tag

kcvan: if i remember correctly i think react div's have `onClick` method so you might be able to use that instead of a `Link` or `a` tag

tc: you shouldn't use div as clickable. it breaks accessible/standards. a tags come with mouse pointer built-in hover states, etc.

kcvan: also to @tommy-bee comment, i think traditionally `a` tags are used to link from page to page
https://www.w3schools.com/tags/tag_a.asp

tc: `button` works as modal opener link too and it comes with the built-in mouse pointer states.

kcvan: +1 for button. i think at this point in time since we dont have a full design for the account page yet we can just leave it as a `Button` component and style it later depending on the design

jerminatorhits: ah i see, TIL. Will refrain from putting `onClick` on divs.

src/pages/Account/AccountPayment.tsx
```diff
@@ -1,10 +1,96 @@
...
+            </Modal>
+          </section>
+        );
+      }}
+    </Query>
+  );
+
+  function handleModal(modal?: ModalType, newPaymentSource?: PaymentSource) {
+    setModal(modal)
+    newPaymentSource && setPaymentSource(newPaymentSource); // null check modal throws undefined error since fade out takes time
```
kcvan: i think we can remove the fadeOut for now and worry about transitions later

src/pages/Account/NewCardForm.tsx
jerminatorhits: sg

src/pages/Account/AccountPayment.tsx
```diff
@@ -61,24 +61,26 @@ const AccountPayment = ({ creditBalance }: Props) => {
 
             <Row>
               <Col xs="12">
-                <div onClick={() => handleModal(ModalType.ADD_NEW_CARD)} className="w-auto d-inline-block align-items-center">
-                  <i className="fas fa-plus-circle" />
-                  <h6 className="ml-2 mb-0 d-inline-block">Add New Card</h6>
-                </div>
+                <a onClick={() => handleModalAction(ModalType.ADD_NEW_CARD)}>
```
tc: doesn't reactstrap button handle icon inside the link for you?

src/styled/customStyles.scss
```diff
@@ -14,4 +14,18 @@
       top: auto;
     }
   }
+}
+
+// bootstrap does not natively have this class
+.cursor-pointer {
```
tc: any reason we need this where not using an `a` or `button` would work since they come with this style built-in?

src/pages/Account/AccountPayment.tsx
jerminatorhits: just to recap:

When `Delete Card` button has been clicked and processed, the modal initiates its fadeOut. However, the `paymentSource` prop provided to the modal is now `undefined` because that card has been deleted. This crashes the app because the modal tries to display `paymentSource.stripeLast4` while fading out.

There are a couple ways to keep it from crashing. (1) have a mock 'placeholder' paymentSource object to fallback to but the modal will look different while its fading out. (2) keep the deleted paymentSource in state (this may cause a memory leak or some sort because technically the card doesn't exist anymore.

Will conditionally render these modals and compromise fadeOut / fadeIn for simplicity sake

src/pages/Account/AccountPayment.tsx
jerminatorhits: hmm are there any examples of this?

src/styled/customStyles.scss
jerminatorhits: discovered they are not built in
https://stackoverflow.com/questions/47269189/bootstrap-v4-css-class-for-pointer-cursor

tc:
```
UPDATE for Bootstrap 4 stable
The cursor: pointer; rule has been restored, so buttons will now by default have the cursor on hover:

<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css">
<button type="button" class="btn btn-success">Sample Button</button>
```

src/pages/Account/AccountPayment.tsx
tc: i added the facebook icon next to the text. The padding can be improved but it does show an icon button.
screen shot 2019-02-20 at 3 08 12 pm

src/styled/customStyles.scss
kcvan: i think we can replace the `Add Card` component with just a `<Button />` component and use  that.  that way we can avoid using an `a` tag for non page links and also get the same functionality

jerminatorhits: replaced with `<Button />` for consistency

src/pages/Account/AccountPayment.tsx
jerminatorhits: thanks. updated

src/pages/Account/AccountPayment.tsx
```diff
@@ -1,10 +1,105 @@
...
+  const [paymentSource, setPaymentSource] = React.useState<PaymentSource | undefined>(undefined);
+  
+  return (
+    <Query query={GET_PAYMENT_SOURCES}>
+      {({ loading, error, data }) => {
+        if (loading) {
+          return <AudioLoading height={48} width={96} />;
+        }
+        if (error || !data) {
+          return <h4>{error ? error.message : 'Error / No Data'}</h4>;
```
tc: We can use an Alert here for the error

src/pages/Account/AccountPayment.tsx
```diff
@@ -1,10 +1,105 @@
...
+        const renderPaymentSources = paymentSources.map((paymentSource: PaymentSource) => (
+          <ListGroupItem
+            className="w-100 d-flex justify-content-between align-items-center"
+            id={paymentSource.id}
+            key={paymentSource.id}>
+            <h6 className="mb-0">
+              {paymentSource.stripeBrand}&nbsp;(...{paymentSource.stripeLast4})
+            </h6>
+            <Button
+              className="btn-icon"
```
tc: can we make this btn-light so it's an action we don't want to emphasize to the user?

src/pages/Account/AccountPayment.tsx
jerminatorhits: sure

src/pages/Account/AccountPayment.tsx
jerminatorhits: 👍

PR #195, opened 2019-02-19
Jeremy/eng 998 security

jerminatorhits:
## Description
Why did you write this code?
Use Front / Reactstrap components

## How to Test
- [ ] Visit '/work/account/security'
- [ ] Verify reset password flow works

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


src/pages/Account/AccountSecurity.tsx
```diff
@@ -1,11 +1,71 @@
...
-      <h1>This is the Security Page</h1>
-    );
+interface AlertProperties {
+  color: string;
+  msg: string;
+  show: boolean;
+}
+
+function AccountSecurity() {
+  const [alert, setAlert] = React.useState<AlertProperties>({ color: '', msg: '', show: false });
```
jerminatorhits: initially had `msg` dictate whether to show or not, but tried to make this component fade away to mimic the snackbar by using the `<Alert transition={...}>` prop, and thought `show` might help achieve fading. This current implementation has the alert with no fade

src/pages/Account/AccountSecurity.tsx
```diff
@@ -1,11 +1,70 @@
...
-    return (
-      <h1>This is the Security Page</h1>
-    );
+interface AlertProperties {
+  color: string;
+  msg: string;
+}
+
+function AccountSecurity() {
+  const [alert, setAlert] = React.useState<AlertProperties>({ color: '', msg: '' });
```
kcvan: do the alerts just pop up in the space that you put it in? i wonder if we can just make it like the native browser alert

src/pages/Account/AccountSecurity.tsx
```diff
@@ -1,11 +1,70 @@
...
+          <ListGroup className="mb-2 d-flex flex-column">
+            {user &&
+              <ListGroupItem
+                className="w-100 d-flex flex-column"
+                disabled={isSubmitting}
+                onClick={() => handleClick(user)}>
+                <h6 className="mb-0">Reset Password</h6>
+                <div className="d-flex justify-content-between">
+                  <h6 className="mb-0 small text-muted">Click here to reset your password</h6>
+                  <i className="fa fa-lock"></i>
```
kcvan: is this the old font awesome?

https://fontawesome.com/icons?d=gallery&q=lock

src/pages/Account/AccountSecurity.tsx
```diff
@@ -1,11 +1,70 @@
...
+interface AlertProperties {
+  color: string;
+  msg: string;
+}
+
+function AccountSecurity() {
+  const [alert, setAlert] = React.useState<AlertProperties>({ color: '', msg: '' });
+  const [isSubmitting, setSubmitting] = React.useState<boolean>(false);
+
+  function handleClick(user: User) {
```
kcvan: i think we can move the handlers below the return now that you're not using `const`

src/pages/Account/AccountSecurity.tsx
jerminatorhits: ty updated. fa -> fas

src/pages/Account/AccountSecurity.tsx
jerminatorhits: 👍 moved

src/pages/Account/AccountSecurity.tsx
jerminatorhits: discussed during standup. will use current behavior for now and consider building native browser alert component when we need it

src/pages/Account/AccountSecurity.tsx
jerminatorhits: update: using `isOpen` prop instead of conditionally rendering the component. Fade should work

src/pages/Account/AccountSecurity.tsx
```diff
@@ -1,10 +1,72 @@
...
+      </Alert>
+
+      <FirebaseConsumer>
+        {({ user }: FirebaseUserProps) => (
+          <ListGroup className="mb-2 d-flex flex-column">
+            {user &&
+              <ListGroupItem
+                className="w-100 d-flex flex-column"
+                disabled={isSubmitting}
+                onClick={() => handleClick(user)}>
```
tc: can `handleClick` be renamed to what it does? is it `handleResetPasswordClick`?

jerminatorhits: 👍

src/pages/Account/AccountSecurity.tsx
```diff
@@ -1,10 +1,68 @@
...
+              </ListGroupItem>
+            }
+          </ListGroup>
+        )}
+      </FirebaseConsumer>
+    </>
+  );
+
+  function handleResetPasswordClick(user: User) {
+    if (!user.email) return;
```
kcvan: do we need some sort of error/redirect if the user's email is missing?

jerminatorhits: this line was mainly to satisfy typescript. Should we redirect to login page if user or user.email does not exist?

tc: prompting an error would work

kcvan: i think if we ran into a case where the user.email doesn't exist maybe we should prompt an error and bring them to the login screen?

jerminatorhits: put in a reactstrap alert for now.

PR #198, opened 2019-02-22
kevin/ENG-1029 (home page)

kcvan:
## Description
Why did you write this code?
Add new brand theme to home page
## How to Test
- [ ] Visit /work
- [ ] Look around

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android


kcvan:
two big changes here:

1) had to add responsive typography based on screen size because bootstrap doesn't have it's own
2) added a fixed height to the header so now we can add the correct sizing/layout based on the header size (4rem)

src/styled/customStyles.scss
```diff
@@ -16,11 +17,200 @@
...
+
+// typography *bootstrap does not come with responsive typography by default
+$custom-font-size-1: 0.875rem;
+$custom-font-size-2: 1.5rem;
+$custom-font-size-3: 2rem;
+$custom-font-size-4: 2.5rem;
+$custom-font-size-5: 3rem;
+$custom-font-size-6: 3.5rem;
+$custom-font-size-7: 4rem;
+$custom-font-size-8: 4.5rem;
```
tc: why 8? 
how are we using these different sizes? headings? 

https://getbootstrap.com/docs/4.3/content/typography/ suggests theres 6 h tags and 4 display headings. are these not sufficient?

src/components/work/Header/Header.tsx
```diff
@@ -39,16 +39,16 @@ const Header = () => {
   const [isOpen, toggleNavbar] = React.useState<boolean>(false);
 
   return (
-    <header className="fixed-top bg-white" id="bee-main-header">
+    <header className="fixed-top bg-white custom-header-height" id="bee-main-header">
```
tc: why can't we use the default header size with default font size provided?

kcvan: the issue here is that there is no default header size, it will shrink/grow to whatever the content is inside the nav.  Since we have a fixed-top nav this creates an issue with any hero/jumbotron.  It makes dealing iwth the header/padding easier if we set a fixed height to the nav so we know how much padding to add on hero/jumbotrons

src/styled/customStyles.scss
kcvan: the issue is that they don't have these classes for multiple screen sizes.  If i use `display-4`, it might be too large on the smallest size device while looking fine on the largest

kcvan: for example, `display-4` looks like this on desktop:
screen shot 2019-02-22 at 12 17 30 pm

but like this on mobile:

screen shot 2019-02-22 at 12 17 34 pm

kcvan: the alternative would be to:
1) create these custom classes
2) create multiple elements and display none on certain sizes

src/components/work/Header/Header.tsx
tc: this page `front-v2.2/html/home/onepage-creative.html#homeSection` looks like it has a fixed header and works on mobile?

src/styled/customStyles.scss
tc: what if you add the responsive rules to the existing classes? `display-4` in this case

src/components/work/Header/Header.tsx
kcvan: this is because they set the padding from the hero content very tall, and both header/hero are with a white background.  then becomes more apparent when you have a blue hero and a white nav, you can see the space between here:

screen shot 2019-02-22 at 12 26 46 pm

kcvan: a few options:
1) set the padding manually on every page
2) the grey background is coming from the jumbotron, we can default the background color to white (this doesnt solve the underlying issue, basically just ignoring the fact that the padding/nav height arent equal

src/pages/home/home.styled.ts
```diff
@@ -0,0 +1,16 @@
+export const CONTENT_CLASSES = {
+  CONTENT_1: {
```
tc: Can you name these what they show? ie `HOST_BENEFITS` etc

src/styled/customStyles.scss
kcvan: removed the classes, will ask Hiroo to add additional media queries when needed

src/pages/home/home.styled.ts
kcvan: sorry that was a placeholder name, i changed it to `FEATURES` based on the sketch file but i can change it to `BENEFITS` if that's what we want

tc: Features work too

src/pages/home/home.config.tsx
```diff
@@ -0,0 +1,91 @@
...
+    svg: <FastCompanySvg />,
+    href: 'https://www.fastcompany.com/40524021/on-this-blockchain-based-version-of-airbnb-theres-no-middleman',
+  },
+  {
+    className: AFFILIATE_CLASSES,
+    svg: <IncSVG />,
+    href: 'https://www.inc.com/darren-heitner/how-this-entrepreneur-is-fixing-250-billion-sharing-economy-with-blockchain-technology.html',
+  },
+  {
+    className: FLEX_CENTER,
```
tc: why is this one different? and why does the className need to be defined in the data array?

src/pages/home/home.config.tsx
```diff
@@ -0,0 +1,91 @@
...
+    src: 'https://placeholdit.imgix.net/~text?txtsize=33&txt=318%C3%97180&w=318&h=180',
+    className: CONTENT_CLASSES.FEATURES.COLUMN_LAYOUT,
+    title: 'Increase Your Income',
+    body:
+      'You will be able to maximize your profit, as our target users are business travelers. Simply list your place, sit back and enjoy.',
+  },
+  {
+    alt: '',
+    src: 'https://placeholdit.imgix.net/~text?txtsize=33&txt=318%C3%97180&w=318&h=180',
+    className: CONTENT_CLASSES.FEATURES.COLUMN_LAYOUT,
```
tc: likewise, here too. Data can be separate from presentation logic.

src/pages/home/home.config.tsx
kcvan: this one is different because it has a slightly different layout on responsive screens being the last item

tc: We should apply the css rule independently of the data still. A programmer won't know that they have to apply this className to the last item.

src/pages/home/home.styled.ts
```diff
@@ -12,6 +12,24 @@ export const CONTENT_CLASSES = {
   }
 };
 
+export const AFFILIATE_CLASSES = `${FLEX_CENTER} mb-4 mb-lg-0`;
+export const TESTIMONIAL_CLASSES = `${FLEX_CENTER} mb-6 mb-lg-0`;
+
+export const AFFILICATION_CONFIG_CLASSES = [
+  AFFILIATE_CLASSES,
```
tc: this and below using fixed sizes is not a good practice. If a programmer adds a data entry in the data array, they have to know to add it here too.

Can you use `length` to generate the classes?

also, is `AFFILICATION` a typo, not sure what this word means

kcvan: im not too sure how to fix this issue then?  it's going to be the same issue as if the programmer adds 4 cards manually and adds the classes on each card, they'll have to know to style the card

kcvan: used a custom class instead of class array

src/styled/customStyles.scss
```diff
@@ -1,5 +1,13 @@
+// home custom styles
+.home-affilications,
```
tc: is affilications suppose to be affiliations?

src/pages/home/home.styled.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const AFFILIATE_CLASSES = `${FLEX_CENTER} mb-4 mb-lg-0`;
+export const TESTIMONIAL_CLASSES = `${FLEX_CENTER} mb-6 mb-lg-0`;
+
+export const JUMBOTRON_STYLES = {
+  backgroundImage: `url('https://static.beenest.com/images/app/misc/painted-ladies2.jpg')`,
+  height: '72vh',
+};
+
+// export const CONTENT_3_STYLES = {
```
tc: remove unused commented code

src/styled/customStyles.scss
kcvan: dont know why my finger keeps hitting the c

src/pages/home/home.styled.ts
kcvan: already did in last commit

PR #203, opened 2019-02-26
Import ES5 autotrack.js

woeltjen:
...to avoid SyntaxError in legacy browsers

https://github.com/googleanalytics/autotrack/issues/137

## Description
Reports of `Uncaught SyntaxError: Use of const in strict mode` when viewing Beenest on legacy browsers (Safari 9/10 were the reports, [Chromium 40](https://www.chromium.org/getting-involved/download-chromium) was any reproduction environment). Generally, these occur when trying to run ES6 code in a browser without ES6 support, such as when an ES6 dependency gets included by webpack without transpilation. In our case, this was the `autotrack` dependency. Ran into miscellaneous errors trying to enable transpilation for `node_modules/autotrack`; instead, changed the `require` call which imports this to point to `autotrack/autotrack.js`, which is a minified ES5 version of the dependency.

## How to Test
1. Build for prod (`npm run prod`)
2. Serve built file (`cd dist; http-server`)
3. Verify that Chromium 40 can load the page

- [x] Chromium on Mac
- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
This solution is not ideal, insofar as it increases the scope of the dependency we include; fine-grained inclusion of plugins would be preferable. Our `index.js` goes from 1575840 bytes to 1582801 bytes with this change. That extra ~7k looks "negligible" at that scale; we should probably look into pruning our other dependencies.

## Learnings
`webpack` can introduce some *obscure* errors...

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -20,15 +20,9 @@ import App from './components/App';
 // Google Analytics to only work on production
 import { AppEnv, APP_ENV } from 'configs/settings';
 if (APP_ENV === AppEnv.PRODUCTION) {
-  require('autotrack/lib/plugins/event-tracker');
-  require('autotrack/lib/plugins/outbound-link-tracker');
-  require('autotrack/lib/plugins/url-change-tracker');
-  require('autotrack/lib/plugins/page-visibility-tracker');
+  require('autotrack/autotrack.js');
   window.ga('create', 'UA-106852049-2', 'auto');
-  window.ga('require', 'eventTracker');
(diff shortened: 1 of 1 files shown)


kcvan:
what

PR #210, opened 2019-02-27
Use sticky positioning for Header

woeltjen:
## Description
This is a proposed simplification of our `Header` to use the `sticky-top` class instead of `fixed-top`

With `fixed-top`, the element no longer influences subsequent flow, so we end up needing to add padding elsewhere any time a `Header` is used (or lose content behind it)

Using `sticky-top` comes out with a similar end result, but the element is treated as part of the flow of the document, so new content comes after the header, only disappearing behind it on scroll.

As an example of the benefit of this, our custom `Jumbotron` wrapper is removed, and `Home` is no longer dependent on padding specified by that custom `Jumbotron` to display correctly.

Note that sticky support is [not universal](https://caniuse.com/#search=sticky), although it is fairly broad among evergreen browsers. Tested in Chromium 40; the header scrolls with the rest of the page, but the layout is otherwise identical.

## How to Test
1. Go to `/work`
2. Verify that page displays as it does on `master`

- [x] Chrome on Mac
- [x] Chromium 40 on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security


kcvan:
i think this is a good fix for any page that scrolls, we'll run into a problem when we have pages that are `100vh` since the header will push the content down below the fold instead of having a full page

PR #213, opened 2019-02-27
Scaffold /work/search

woeltjen:
## Description
Begin implementing the subset of the new Search user interface that is not expected to change as a consequence of user testing

## How to Test
1. Go to /work/search
2. Expect to see a search page with same layout as the design

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. Integration with front page search box
2. Implementation of additional "new search" features


src/pages/search/SearchForm.tsx
```diff
@@ -0,0 +1,39 @@
...
+    label: 'Price Range',
+    component: <strong>Price range!</strong>
+  },
+  {
+    label: 'Transit Time',
+    component: <strong>Transit time!</strong>
+  },
+  {
+    label: 'More Filters',
+    component: <strong>Moar filters!</strong>
```
kcvan: how many moar, doe

src/pages/search/SearchFilter.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface Props {
+  children: React.ReactNode;
+  label: String;
+}
+
+const SearchFilter = ({ children, label }: Props) => {
+  const [isOpen, setOpen] = React.useState<boolean>(false);
+  return (
+    <Dropdown isOpen={isOpen} toggle={() => setOpen(!isOpen)}>
```
kcvan: do you know how easy it is to drop in form items like check boxes and stuff like that in a dropdown?  I havent tried it yet but I was also looking at the `<Collapse />` component for the filter menus that might be easy to use

woeltjen: `Dropdown` seems to accept arbitrary content reasonably well. I kinda like the `Collapse` approach from a UX perspective though, it makes room for the form filters without obscuring any other content.

In any event, I'll be hiding these controls before this PR is over (issue says "Ignore filters for now" which I have already failed to do) but let's try to keep that in mind when we do get there

woeltjen:
Marking this ready for review. Currently displays test data and no links have been implemented; figure we can focus on display-related concerns for now to make reviews simpler.

Note the use of `style` attributes. Tried to avoid these, but ran into a couple of cases where I couldn't find an idiomatic bootstrap solution. Suggestions for cleaner solutions are welcome!

src/components/work/Header/Header.tsx
```diff
@@ -39,7 +39,7 @@ const Header = () => {
   const [isOpen, toggleNavbar] = React.useState<boolean>(false);
 
   return (
-    <header className="sticky-top bg-white" id="bee-main-header">
+    <header className="sticky-top bg-white" id="bee-main-header" style={{ 'minHeight': '4.125rem' }}>
```
woeltjen: Here, a gap was sometimes showing up between the Header and SearchBar in mobile; root cause is that the header changes in height slightly. Adding an explicit `min-height` allows me to assume a constant height for other sticky elements elsewhere. (Using `position: sticky` instead of `position: fixed` was supposed to allow me to ignore that, but as soon as I need other `sticky` elements it becomes a concern again.)

tc:
screen shot 2019-02-27 at 4 21 01 pm

i think the search bar was intended to be less than 100%
screen shot 2019-02-27 at 4 21 56 pm

src/components/work/Header/Header.tsx
kcvan: i think we have a `custom-header-height` class in the customStyles class we can use

kcvan: you can probably change the 4rem to `4.125rem` if you need it to be that large

src/pages/search/ListingCard.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ListingShort } from 'networking/listings';
+
+const ListingCard = ({
+  homeType,
+  listingPicUrl,
+  pricePerNightUsd,
+  sleepingArrangement,
+  title
+}: ListingShort) => (
+  <Card>
```
kcvan: we can add `tag={Fade}` and import `Fade` from reactstrap to make these cards fade in so they dont appear abruptly

src/components/work/Header/Header.tsx
woeltjen: That. Helps. So. Much.

🙏

src/pages/search/Search.tsx
```diff
@@ -0,0 +1,164 @@
...
+    prices: [],
+    sleepingArrangement: '2 Beds',
+    state: 'CA',
+    title: 'Actual nest of beeeeeeeees'
+  }
+];
+
+const Search = () => {
+  return (
+    <>
```
kcvan: i think for a lot of our fragments we can use the `<Fade>` component instead now to have fade-in page transitions

tc:
thanks for removing the inline css

woeltjen:
> thanks for removing the inline css

Actually, there's still some in Search. (Got one of the three `style` usages sorted out with help from @kcvan)

For the remaining two, should I add `customStyles` for these? I'd definitely prefer to find a way to get sensible `sticky` behavior for header/searchbar/map using bootstrap built-ins, but not seeing a way. (Or am I getting ahead of myself in assuming these components should even have that display behavior?)

tc:
@woeltjen can it just be fixed?

woeltjen:
> @woeltjen can it just be fixed? or is sticky === fixed in css terms?

You run into the same problem with `fixed` (need to do some kind of ad hoc positioning/z-indexing to avoid losing content behind the `fixed` element).

I'll try moving into `customStyles`, think the remaining in-line styles can be refactored down to something reasonable there.

woeltjen:
@tc Have it down to four CSS properties across two classes in `customStyles` with 1a4f1a3. Not seeing a way to reduce custom CSS further without changing behavior, but open to suggestions

tc:
@woeltjen i think that's fine

PR #214, opened 2019-02-27
add profile photo uploader to work/account

jerminatorhits:
## Description
Why did you write this code?
add profile photo uploading to account page

## How to Test
- [ ] Visit work/account
- [ ] upload photo successfully

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?
The old site straight up does not display this component on Mobile and Tablet
Right now, it is displayed for all tabs (it'll just kick down to the bottom of screen).
This might be kind of noisy for the user to always have it there. Might consider making another tab separately for it or wait on designs for other alternatives.

## Learnings

### DIFF
--- a/src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
+++ b/src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
@@ -43,9 +43,9 @@ const ProfilePhotoUploaderContainer = styled.div`
     }
     .hide-uploader {
       .dz-message {
-          display: flex !important;
+        display: flex !important;
       }
-      .dz-previow {
+      .dz-preview {
         display: none !important;
       }
     }
--- a/src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
+++ b/src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
@@ -1,0 +1,64 @@
+import styled from 'styled-components';
+import { color, typography } from 'styled/utils';
+
+const ProfilePhotoUploaderContainer = styled.div`
+  .photo-uploader-container {
+    background-size: cover;
+    background-position: center;
+    border-radius: 50%;
+    display: block;
+    overflow: hidden;
+    position: relative;
+    &:after {
+      border-radius: 100%;
+      box-shadow: 0px 0px 0px 2000px #E3DFD2;
+      content: '';
+      height: 240px;
+      left: 0px;
+      pointer-events: none;
+      position: absolute;
+      top: 0px;
+      width: 240px;
+    }
+    &:hover {
+      .dz-message {
+          background-color: ${color('body', 0.9)};
+      }
+    }
+    .dz-message {
+      ${typography('title', 9)};
+      background-color: ${color('body', 0.6)};
+      bottom: 0px;
+      color: white;
+      flex-direction: row-reverse;
+      height: 58px;
+      margin: 0 !important;
+      padding: 13px 22px 21px;
+      position: absolute;
+      transition: all .15s ease-in-out;
+      width: 207px;
+      .bee-svg {
+          margin-left: 8px;
+      }
+    }
+    .hide-uploader {
+      .dz-message {
+          display: flex !important;
+      }
+      .dz-previow {
+        display: none !important;
+      }
+    }
+    .filepicker {
+      background-color: ${color('body', 0)} !important;
+      border: none !important;
+      border-radius: 50% !important;
+      height: 240px !important;
+      transition: all .2s ease-in-out !important;
+      width: 240px !important;
+      transition: all .2s ease-in-out;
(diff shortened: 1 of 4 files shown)


kcvan:
have re figured out a way to upload it without having a refresh?

src/pages/Account/Account.tsx
```diff
@@ -82,6 +82,11 @@ const Account = () => {
                     <Route component={NotFound} />
                   </Switch>
                 </Col>
+                <Col md={1} />
```
kcvan: if it's only one parameter you can do `md="1"`, i think if you just do `{1}` it expects { size: '', etc: 'etc' }

jerminatorhits: ahh gotcha.

also, I don't actually know when to use md, sm, lg, and if when it is good to be specific. I want to declare the column size based on 12 and so I stick to `md` for that declaration. ¯\_(ツ)_/¯

src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
```diff
@@ -0,0 +1,64 @@
+import styled from 'styled-components';
+import { color, typography } from 'styled/utils';
+
+const ProfilePhotoUploaderContainer = styled.div`
+  .photo-uploader-container {
```
tc: i assume this is a copy and paste from old component?

jerminatorhits: mhm

tc: it's ok it is a copy/pasta but if it's completely unchanged, can't we just use the existing component?

jerminatorhits: ooo the old one has media queries where there's a display none when the screen is less than desktop. In this implementation in an effort to shy away from media queries, the photo uploader is always shown and I removed that display none property.

If the old component was used, there would be an empty `<Card />` component (it wraps ProfilePhotoUploader) on Tablet and Mobile.

jerminatorhits: image

tc: ok, ideally we want to stick to the components front already provided for us to avoid re-writing CSS.
see:
screen shot 2019-02-27 at 3 22 51 pm

The photo uploader can just be a button "Change Photo" vs drag and drop to save effort.

jerminatorhits:
@kcvan The photo updates for me without having to refresh

src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
jerminatorhits: Currently, the networking to upload to our backend and S3 is happening in the `<PhotoUploader />` component. Should I try to route an input field through that component or perhaps make a new gql endpoint just for photo uploading?

tc: can we change the styling without changing the functionality?

kcvan: i think we can do the same thing we're doing in the old header where it checks the db, firebase, and then returns the default photo.  that way, we can ensure we get an updated photo once the DB updates

src/pages/Account/Account.tsx
kcvan: its the same as media queries, you can think of sm = mobile, md = tablet, lg = desktop, xl = very large.  whatever you specify as the smallest will propagate up to largest screen.  usually if you're on mobile, if you're inside a row it normally has `flex-wrap` if i remember correctly youll have to do something like `sm="12"` if you want the entire column to span the width of the screen, or else it'll stack and then wrap.

src/components/work/ProfilePhotoUploader/ProfilePhotoUploader.container.ts
jerminatorhits: don't think so since the styles are coupled within the PhotoUploader Component, I can try to recreate new component: the Dropzone component that use bootstrap styles only?

tc: let's just hold off on it for now. this PR looks fine with the current profile upload design.

jerminatorhits: ok sg

PR #222, opened 2019-02-28
Jeremy/eng 997b current upcoming

jerminatorhits:
## Description
Why did you write this code?
display current and upcoming trips
port over the contact host modal to bootstrap
port over the cancelBooking alert/modal to bootstrap

## How to Test
- [ ] Visit '/work/trips'
- [ ] Click 'upcoming' and 'current' to see respective bookings
- [ ] Send message via 'Contact Host' to host successfully
- [ ] Successfully cancel a booking using 'Cancel Booking'

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


jerminatorhits:
Note: Using the same cards for all 4 tabs. Weird thing was that I was able to cancel a booking on the 'Cancelled/Rejected' tab.

Next PR will do the following:

- remove 'cancel booking' button
- add the red status text for cancelled/reject cards
- fix photo image to fit different screens and card sizes, without quality loss.

src/components/work/CancelBookingModal.tsx/CancelBookingModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  if (isSubmitting) {
+    return <LoadingPortal currency={currency} />;
+  }
+
+  return (
+    <Modal isOpen toggle={() => !isSubmitting && handleModalAction}>
+      <ModalHeader>Cancel Booking</ModalHeader>
+      <ModalBody>
+        <h6>Are you sure you want to cancel this booking?</h6>
+        {booking && <h6>{`Booking: ${booking.id}`}</h6>}
```
woeltjen: This looks equivalent to:

    <h6>Booking: {booking.id}</h6>

(`booking` is non-optional in Props so the check has been handled at compile-time, and there's some extra syntax around the templated part)

jerminatorhits: oo makes sense. thanks!

src/components/work/CancelBookingModal.tsx/CancelBookingModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </ModalBody>
+      <ModalFooter>
+        <Button color="secondary" disabled={isSubmitting} onClick={() => handleModalAction()}>Back</Button>{' '}
+        <Button color="danger" disabled={isSubmitting} onClick={() => handleCancelBooking()}>Yes, Cancel Booking</Button>
```
woeltjen: Both `onClick` handlers are no-argument functions so you should be able to pass them in directly:

    onClick={handleModalAction}
    onClick={handleCancelBooking}

jerminatorhits: brilliant. will keep in mind in the future!

src/components/work/CancelBookingModal.tsx/CancelBookingModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </ModalBody>
+      <ModalFooter>
+        <Button color="secondary" disabled={isSubmitting} onClick={handleModalAction}>Back</Button>{' '}
+        <Button color="danger" disabled={isSubmitting} onClick={handleCancelBooking}>Yes, Cancel Booking</Button>
+      </ModalFooter>
+    </Modal>
+  );
+
+  function handleCancelBooking() {
+    if (!booking) return;
```
woeltjen: This check is also unnecessary (handled at compile-time by enforcing Props interface)

jerminatorhits: 👍  the node backend in me must've gotten nervous =x

src/components/work/CancelBookingModal.tsx/CancelBookingModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { differenceInDays } from 'date-fns';
+import { cancel, loadWeb3 } from 'utils/web3';
+import { AlertProperties } from 'components/work/Alert/Alert';
+import LoadingPortal from 'components/work/LoadingPortal';
+import { getFriendlyErrorMessage } from 'utils/validators';
+
+interface Props {
+  booking: Booking;
+  cancelBooking: (booking: Booking) => Promise<Booking>;
+  handleModalAction: () => void;
```
woeltjen: For consistency with React naming conventions, I'd expect this to be `onModalAction` at the interface level and `handleModalAction` at the implementation level, for usage like:

    <CancelBookingModal onModalAction={handleModalAction} etc... />

src/components/work/ContactHostFormModal/ContactHostFormModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  [ContactUserField.SUBJECT]: '',
+  [ContactUserField.MESSAGE]: '',
+};
+
+const ContactHostSchema = Yup.object({
+  subject: Yup.string().required('Please fill out the subject field.'),
+  message: Yup.string().required('Please fill out the message field.'),
+});
+
+function ContactHostForm({ booking, contactUser, handleModalAction }: Props) {
```
woeltjen: Oh yay, this will come in handy for the Contact Host button on the listing page

src/components/work/ContactHostFormModal/ContactHostFormModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { CONTACT_USER, ContactUserField, User } from 'networking/users';
+import { Button, Form, FormGroup, Label, FormFeedback, Input, ModalFooter, ModalBody, Modal, ModalHeader } from 'reactstrap';
+import Textarea from 'components/shared/Textarea';
+import { TextareaEvent } from 'components/shared/Textarea/Textarea';
+import Loading from 'components/shared/loading/Loading';
+
+interface Props {
+  contactUser: (input: ContactUserInput) => Promise<EmailResponse>;
+  booking: Booking;
+  handleModalAction: () => void;
```
woeltjen: As above, I'd expect this to be named `onModalAction`

src/components/work/ContactHostFormModal/ContactHostFormModal.tsx
jerminatorhits: 🎉

src/components/work/CancelBookingModal.tsx/CancelBookingModal.tsx
jerminatorhits: that was bugging me in the back of my head. I'll try to remember or refer back to this in the future. tyty!

src/components/work/ContactHostFormModal/ContactHostFormModal.tsx
jerminatorhits: 👍 done

src/components/work/TripCard/TripCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <div className="embed-responsive embed-responsive-4by3">
+        <div className="embed-responsive-item">
+          <CardImg className="h-100" top src={booking.listing.listingPicUrl} alt="Listing Cover Photo" />
+        </div>
+      </div>
+      <CardBody>
+        <CardTitle className="h5 font-weight-normal mb-3">{booking.listing.title}</CardTitle>
+        <CardSubtitle className="small mb-3">
+          {addressLine1 && formatAddress(addressLine1, addressLine2, city, state, country, postalCode)}
+          {!addressLine1 && formatGeolocationAddress({ lat, lng, city, country })}
```
woeltjen: Small change, but I'd do this as:

    {addressLine1 ? formatAddress(...) : formatGeolocationAddress(...)}

Rationale is maintainability; the logic for the two checks is necessarily complementary (we want exactly one of these formatted outputs to be displayed) but separating them allows them to diverge (e.g if we change `addressLine1` to `(addressLine1 || addressLine2)` on one line but not the next)

Also, I may be misremembering, but I think we added a `fullAddress` field to listings on the backend that handles this formatting ahead of time, so you might just be able to say `booking.listing.fullAddress` here.

woeltjen: Yep, backend defines `fullAddress` as:

      fullAddress: {
        type: DataTypes.VIRTUAL,
        get: function () {
          const { addressLine1, addressLine2, city, state, country, postalCode, lat, lng } = this;
          return addressLine1 ?
            formatAddress(addressLine1, addressLine2, city, state, country, postalCode) :
            formatGeolocationAddress({ lat, lng, city, country });
        }
      }

May need to update GraphQL queries to expose it tho; in the GraphQL `booking` end point there is a `Listing` type defined which does not include this.

jerminatorhits: interesting. definitely on board with the maintainability factor (leaves smaller margin of error as stated)

As for the `fullAddress`, I think some cards should only show city, state, country, etc. (omitting street address since it shouldn't be shown to unapproved bookings). Would it be better practice to handle that in the backend?

jerminatorhits: actually lemme smoke test this

woeltjen: > As for the `fullAddress`, I think some cards should only show city, state, country, etc. (omitting street address since it shouldn't be shown to unapproved bookings). Would it be better practice to handle that in the backend?

Ah, you're right; `fullAddress` (as currently defined) should only be exposed for approved bookings. Sounds like we want something like a `displayAddress` that takes into account the privileges of the user doing the viewing? In any case, that's getting out of scope for this issue.

jerminatorhits: ahh makes sense. Will put in the fix for a ternary at the very least

tc:
The Contact Host  should be an `a` tag or `button` so it gets a natural mouse cursor

src/pages/trips/Trips.tsx
```diff
@@ -1,37 +1,97 @@
...
-        // const isUpcomingEmpty = !(upcoming || []).length;
+        const isUpcomingEmpty = !(upcoming || []).length;
+        const renderCurrentEmpty = 
+          <>
+            <Row className="mb-2">
+              <h2>You have no trips awaiting approval.</h2>
+            </Row>
+            <Row>
+              <a href='/work'>
+                <Button>Book a Home Today!</Button>
```
tc: this can be `<a className="btn">` so the cursor is correctly registered.

jerminatorhits: 👍

jerminatorhits:
seems like `<a>` tag only provides the pointer when an `href` is declared (which won't work for the modal).

`Button` has a lot of intrinsic padding as well
image

Made a custom css class after doing some research on a pretty back-and-forth discussion: https://github.com/twbs/bootstrap/issues/23709

```
.c-pointer {
 cursor: pointer;
}
```

tc:
@jerminatorhits curious, could you add `href="#"`  to the a tag? I think the javascript does a `preventDefault()` to overridden the call.

src/components/work/TripCard/TripCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <CardText>Status: {displayStatus}</CardText>
+        <CardText className="small">
+          Booking: <span>{id}</span>
+        </CardText>
+      </CardBody>
+      <CardFooter className="text-center">
+        <Row className="align-items-center">
+          <Col
+            xs="4"
+            onClick={() => handleModalAction(ModalType.CONTACT_HOST)}
```
tc: could this be `onClick={handleOpenContactHostModal}` and the method call `handleModalAction(ModalType.CONTACT_HOST)` ?

jerminatorhits:
@tc put an `href="#"` on the `a` tag and it has the pointer but goes to home page.

src/components/work/TripCard/TripCard.tsx
jerminatorhits: could do. since there's also a `CANCEL_BOOKING` modal, that would require a `handleOpenCancelBookingModal` to be passed into this component as well. If it's conventionally better to use specific handle names, I'm all for it.

tc: i think using a method can ensure better enforcement vs a parameter.

tc:
``@tc put an href="#" on the a tag and it has the pointer but goes to home page.``
does the onClick listener have a `e.preventDefault` or `e.stopPropagation` ?

src/components/work/TripCard/TripCard.tsx
jerminatorhits: i see. so the component doesn't have to know about the enum. will change

jerminatorhits:
@tc `<a href="#" onClick={(event: React.MouseEvent<HTMLElement>) => event.preventDefault()}>`

includes the pointer and opens the modal. Interesting!

PR #224, opened 2019-03-01
Wire SearchBar to Search page

woeltjen:
## Description
Integrate functionality into new-theme Search page

## How to Test
1. Go to `/work`
2. Search it up!
3. Expect to see search results presented in the new theme

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Search filters?

### DIFF
--- a/src/components/work/SearchBar/SearchBar.tsx
+++ b/src/components/work/SearchBar/SearchBar.tsx
@@ -173,7 +173,7 @@ class SearchBar extends React.Component<RouterProps, State> {
     event.preventDefault();
     const { bounds, coordinates, checkInDate, checkOutDate, numberOfGuests } = this.state;
     const locationQuery = this.inputRef.current ? this.inputRef.current.value : '';
-    return window.location.href = `${BEENEST_HOST}/listings?${stringifyQueryString({
+    return window.location.href = `${BEENEST_HOST}/work/search?${stringifyQueryString({
       locationQuery,
       utm_term: locationQuery,
       ...(bounds && { bounds }),
--- a/src/networking/listings.ts
+++ b/src/networking/listings.ts
@@ -174,6 +174,7 @@ const LISTING_CARD_FRAGMENT = gql`
   fragment ListingCard on Listing {
     city
     country
+    homeType
     id
     idSlug
     pricePerNightUsd
@@ -182,6 +183,7 @@ const LISTING_CARD_FRAGMENT = gql`
       pricePerNight
       securityDeposit
     }
+    sleepingArrangement
     state
     title
   }
@@ -200,7 +202,6 @@ const LISTING_DETAILS_FRAGMENT = gql`
     checkOutDate
     checkOutTime
     description
-    homeType
     houseRules
     lat
     lng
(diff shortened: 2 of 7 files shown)


kcvan:
looks great, vic!  I'm not sure if we want to do this now, but i think the cards need to have a fixed height so we dont get variable heights like this: 
screen shot 2019-02-28 at 7 47 40 pm

im also seeing this on mobile/tablet:
screen shot 2019-02-28 at 7 47 40 pm
screen shot 2019-02-28 at 7 47 50 pm


i think the ui is passable on tablet because of the larger screen size, but on mobile it covers almost the entire page.  it also doesn't allow the header to drop down when clicking on the hamburger (i think it's chilling behind the search bar).  maybe we can collapse the search bar for mobile/tablet?

tc:
Is that a dot or period above the listing title?

woeltjen:
> I'm not sure if we want to do this now, but i think the cards need to have a fixed height so we dont get variable heights like this:

Yeah, I spent a bit of time trying to get that to work, but ultimately decided to punt. `CardDeck` will normalize heights, but getting a consistent or desired number of columns displayed started getting tricky with that. Having a specified height in rem is probably fine.

> Is that a dot or period above the listing title?

Good catch! It's the dot that's supposed to be between `homeType` and `sleepingArrangement`, but the GraphQL query for search wasn't asking for those. Updated, and will do a backend PR to make sure Agoda listings expose that data.

woeltjen:
@kcvan w.r.t. the second point about search bar, the simplest fix I could find was to remove `sticky` from the search bar's container. It scrolls away on mobile (sounds like what we want) and also scrolls away on desktop, tho the map still stick(ie)s around. Also removed z-indexing on that component so hamburger menu is no longer blocked (it is see-through though; think it just needs a `bg-white` but think we should handle that in a separate PR).

jerminatorhits:
The Card Height issue was one I came across for the Trips page. Not sure if the following violates good bootstrap classes but here's what I came up with:
```
<Container>
  <Row>
    {data[category].map((booking: Booking) => (
      <Col key={booking.id} className="d-flex" md="6" lg="4">
        <TripCard
          key={booking.id}
          category={category}
          booking={booking}
          handleModalAction={(modal: ModalType) => handleModalAction(modal, booking)} />
     </Col>
    ))}
  </Row>
</Container>
```
will continue trying different things to improve styles in the meantime

woeltjen:
@jerminatorhits Thanks, using `d-flex` worked perfectly!

kcvan:
i remember running into the invisible hamburger menu issue before, `sticky-top` actually fixed it but if we're running into it again i think there's a different underlying problem than `bg-white`.  The reason i think this is because when you open the header, the svg shifts up a little bit which is strange.

tc:
screen shot 2019-03-01 at 3 24 50 pm

why do some say 1 room but some say "2"

woeltjen:
Same reason we see "Sleeping arrangement: 2" for those listings.

image

We ask users to enter number and type of bed, but we don't validate that they've done anything more than provided text.

tc:
i see, maybe just remove the number and just leave in "Entire Place" until we can normalize the data for clarity.

PR #229, opened 2019-03-01
Use fixed-top for Header

woeltjen:
## Description
@kcvan noted that the switch to `sticky-top` from #210 was causing the hamburger menu to push down content on tablet/mobile.

This proposed change goes back to `fixed-top`, so that the menu appears over content instead of pushing it down, and additionally adds a `main` tag with a `bee-top` class surrounding the top-level `Switch`, so that individual pages do not need to be concerned with this height (achieving the goal of #210 a different way)

## How to Test
1. View /work on mobile
2. Click hamburger menu
3. Expect content below to stay in place and get covered by menu

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Indirectly related: Should we bring in `Footer`?

### DIFF
--- a/src/components/work/Header/Header.tsx
+++ b/src/components/work/Header/Header.tsx
@@ -51,7 +51,7 @@ const DetailedHeader = () => {
   const [isOpen, toggleNavbar] = React.useState<boolean>(false);
 
   return (
-    <header className="sticky-top bg-white custom-header-height shadow" id="bee-main-header">
+    <header className="fixed-top bg-white custom-header-height shadow" id="bee-main-header">
       <Navbar light expand="md">
         <NavbarBrand href="/work">
           <BeenestSVGPrimary />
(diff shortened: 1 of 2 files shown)


kcvan:
i think this change is going to affect the first section of every page, we might need a 
```
.bee-header > * {
  padding-top: 4.125rem;
}
```

or something to accommodate for the header.

if we go this route, we'll also need to add class/id names on the pages that we don't want the extra padding

kcvan:
sorry read the last part of your comment, disregard my previous comment

kcvan:
im half afraid to approve this now that we've been using `sticky-top` for so long and built things around the header being sticky-top

tc:
@woeltjen @kcvan what's the update on this one now?

woeltjen:
> @woeltjen @kcvan what's the update on this one now?

Looks like we don't want, closing

PR #234, opened 2019-03-02
Lay out /work/listings/:id

woeltjen:
## Description
Lay out the contents of the Listing page.

## How to Test
1. Navigate to /work/listings/95_vic (or some other listing)
2. Verify that major page elements are laid out sensibly
    1. Image gallery at top
    2. Listing details below (to left, in desktop, or at center, on tablet/mobile)
    3. Booking Card to the right (or at bottom, on tablet/mobile)

Note that individual page elements are still unstyled (subsequent PRs!)

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. Gallery layout (move button)
2. View Photos functionality
3. Contact Host functionality
4. Booking Card layout
5. Booking Card functionality

### DIFF
--- a/src/pages/listing/ListingPage.tsx
+++ b/src/pages/listing/ListingPage.tsx
@@ -4,6 +4,7 @@ import { Button, Col, Container, Row } from 'reactstrap';
 import ListingGallery from './ListingGallery';
 import ListingInformation from './ListingInformation';
 import BookingCard from './BookingCard';
+import BookingBar from './BookingBar';
 
 import { Listing } from 'networking/listings';
 
@@ -22,12 +23,9 @@ const ListingPage = (listing: Listing) => (
         </Col>
       </Row>
     </Container>
-    <Row className="fixed-bottom bg-white d-lg-none w-100 p-2 align-items-center justify-content-between">
-      <Col>
-        ${listing.pricePerNightUsd} per night
-      </Col>
-      <Button>Request to Book</Button>
-    </Row>
+    <footer className="fixed-bottom bg-white d-lg-none">
+      <BookingBar {...listing} />
+    </footer>
   </Row>
 );
 export default ListingPage;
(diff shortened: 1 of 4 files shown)


woeltjen:
This seems like it's in a reasonable state to review. One thing worth noting is that the BookingBar will obscure part of the footer on tablet/mobile. This is actually the case for the old design, too, so this is at parity, but raising it in case anyone sees a solution

src/pages/listing/ListingPage.tsx
```diff
@@ -1,14 +1,31 @@
...
 
 import { Listing } from 'networking/listings';
 
-const ListingPage = (listing: Listing) => (<>
-  <ListingGallery {...listing} />
-  <ListingInformation {...listing} />
-  <BookingCard {...listing} />
-</>);
+const ListingPage = (listing: Listing) => (
+  <>
```
kcvan: I think this can be <Fade>

tc: @kcvan  your comment got cut off

kcvan: ah thanks, was going to say this can be `<Fade>`

src/pages/listing/ListingPage.tsx
```diff
@@ -9,7 +9,7 @@ import BookingBar from './BookingBar';
 import { Listing } from 'networking/listings';
 
 const ListingPage = (listing: Listing) => (
-  <>
+  <Fade>
```
tc: sometimes it's not good to use a `Fade` effect. I think it's fine here.

ie, if we want the page to load immediately, like the signup/booking pages. Anything slowing the user down to conversion is not good. cc @kcvan

PR #241, opened 2019-03-04
Jeremy/eng 1062 work receipt

jerminatorhits:
## Description
Why did you write this code?
rebrand receipt page

## How to Test
- [ ] Visit work/trips/:id/receipt
- [ ] View different screen widths

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/utils/formatter.ts
+++ b/src/utils/formatter.ts
@@ -2,9 +2,9 @@ import format from 'date-fns/format';
 import { Listing } from "networking/listings";
 
 // comma-separates terms and leaves no trailing commas
-export function formatAddress(...args: Array<string>): string {
+export function formatAddress(...args: Array<string | undefined>): string {
   const clean = args.filter(str => !!str && !(str.toUpperCase() === 'US' || str.toUpperCase() === 'USA'));
-  return /^([0-9]|-)+$/.test(clean[clean.length - 1]) // No comma before postal code, if present
+  return /^([0-9]|-)+$/.test(clean[clean.length - 1] || '') // No comma before postal code, if present
     ? clean.slice(0, -1).join(', ') + ' ' + clean.slice(-1)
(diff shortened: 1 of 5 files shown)


src/utils/formatter.ts
```diff
@@ -4,6 +4,9 @@ import { Listing } from "networking/listings";
 // comma-separates terms and leaves no trailing commas
 export function formatAddress(...args: Array<string | undefined>): string {
   const clean = args.filter(str => !!str && !(str.toUpperCase() === 'US' || str.toUpperCase() === 'USA'));
+  if (clean.length < 3) {
+    return clean.join(', ');
+  }
```
jerminatorhits: noticed some short addresses displayed as `SAN FRANCISCO CA` with no comma. This may be a band-aid for those small cases.

also not sure if `Vancouver, British Columbia, CAN` or `Vancouver, British Columbia CAN` is correct

woeltjen: The way I tried to deal with this on the backend was:

    const clean = args.filter(str => !!str && !(str.toUpperCase() === 'US' || str.toUpperCase() === 'USA'));
    return /^([0-9]|-)+$/.test(clean[clean.length - 1]) ? // No comma before postal code, if present
      clean.slice(0, -1).join(', ') + ' ' + clean.slice(-1) :
      clean.join(', ');

...to use a regex to see if the last element looked like a postal code.

jerminatorhits: wow nice. will use thx!

PR #244, opened 2019-03-04
Move map behind date picker

woeltjen:
## Description
Map on search page was covering date range inputs; this changes z-indexing so it does not.

## How to Test
1. Search for some listings
2. Open the date range picker on the search page
3. Verify that control is not blocked by map
4. Scroll down
5. Verify that map disappears *under* header, not over

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/styled/customStyles.scss
+++ b/src/styled/customStyles.scss
@@ -54,11 +54,11 @@ $HEADER-HEIGHT: 4.125rem;
 
 .bee-top {
   top: 4.125rem;
-  z-index: 999; // Behind bootstrap overlays
 }
 
 .bee-search-map {
   height: calc(100vh - 11.25rem);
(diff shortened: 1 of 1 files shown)


src/styled/customStyles.scss
```diff
@@ -54,11 +54,11 @@ $HEADER-HEIGHT: 4.125rem;
 
 .bee-top {
```
kcvan: i think we can use `$HEADER-HEIGHT` here

PR #252, opened 2019-03-05
add page if /work/search is empty

kcvan:
## Description
Why did you write this code?
feat(search): add error mage if search is empty
## How to Test
- [ ] Visit /work/search
- [ ] Verify it asks you to enter a city or region and search bar shows up

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

### DIFF
--- a/src/pages/search/Search.tsx
+++ b/src/pages/search/Search.tsx
@@ -5,6 +5,7 @@ import { Query } from 'react-apollo';
 import { SEARCH_LISTINGS } from 'networking/listings';
 import Footer from 'components/work/Footer';
 import Loading from 'shared/loading/Loading';
+import { getFriendlyErrorMessage } from 'utils/validators';
 
 import SearchBar from 'components/work/SearchBar';
 import { LISTING_CARD_IMAGE_DIMENSIONS } from 'utils/imageDimensions';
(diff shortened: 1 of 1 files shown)


src/pages/search/Search.tsx
```diff
@@ -29,11 +30,38 @@ const Search = () => {
...
+              <Row className="px-4" noGutters>
+                <h5 className="text-center">Please enter in a city or region to see results.</h5>
+              </Row>
+            </Container>
+          );
+        } else if (error) {
+          return (
+            <Container tag={Fade} className={`${VIEWPORT_CENTER_LAYOUT} flex-column bee-without-header-height-container`}>
+              <h2 className="font-size-md-down-5 text-center">Error: {error.message}</h2>
+              <h5 className="text-center">Please contact support if this continues.</h5>
```
tc: this should have a link

src/pages/search/Search.tsx
```diff
@@ -29,11 +30,38 @@ const Search = () => {
...
-      }
+      {({ loading, error, data }) => {
+        if (loading) {
+          return (
+            <Container tag={Fade} className={VIEWPORT_CENTER_LAYOUT}>
+              <Loading height="8rem" width="8rem" />
+            </Container>
+          );
+        }
+        if (error && error.message.includes('400') && Object.keys(queryParams).length === 0) {
```
tc: this is bad: `error.message.includes('400')`

tc: apollo errors has an `invalidArgs` parameter you can use.

tc: screen shot 2019-03-05 at 1 10 42 pm

i'm not sure this works either. I did an empty search and got an error instead of the search bar.

kcvan: ah it's because there's a 1 for guests, ill add a check for that

kcvan:
to be used with backend PR: https://github.com/thebeetoken/beenest-backend/pull/795

src/pages/search/Search.tsx
kcvan: 👍

woeltjen:
When I go to `/work/search` I get:

image

Running with `kevin/ENG-1072` on the backend

src/pages/search/Search.tsx
```diff
@@ -29,11 +31,65 @@ const Search = () => {
...
-        loading ? <Loading /> :
-        error ? <h2>Error: {error.message}</h2> :
-        <SearchPage listings={data.searchListings} />
-      }
+      {({ loading, error, data }) => {
+        if (loading) {
+          return (
+            <Container tag={Fade} className={VIEWPORT_CENTER_LAYOUT}>
+              <Loading height="8rem" width="8rem" />
+            </Container>
```
woeltjen: Oh yay, having the little loading thing stuck in the top-left was bugging me!

Can this be refactored into its own component, e.g. `LoadingPage`? That way we can just drop in an `if (loading) return <LoadingPage />;` wherever we GraphQL

src/pages/search/Search.tsx
```diff
@@ -29,11 +31,65 @@ const Search = () => {
...
+              </Row>
+              <Row className="d-flex justify-content-center px-0 mx-0 bg-white bee-top">
+                <Col tag={Fade} className="w-100 p-5">
+                  <SearchBar />
+                </Col>
+              </Row>
+            </div>
+          </Container>
+          );
+        }
```
woeltjen: Corollary to the above, could these be their own components? I'm picturing something approximately like:

    if (loading) {
      return <LoadingPage />;
    }
    if (error) {
      return isBadUserInputError(error) ? <SearchErrorPage /> : <ErrorPage error={error} />
    }
    if (searchListings.length < 1) {
      return <EmptySearchPage />;
    }
    return <SearchPage listing={searchListings} />;

I've been leaning on this as a more general pattern when using `Query`, as deep nesting can get hard to follow.

src/pages/search/Search.tsx
```diff
@@ -29,11 +31,65 @@ const Search = () => {
...
+
+        const { searchListings } = data;
+        if (searchListings.length === 0) {
+          return (
+            <Container tag={Fade} className="d-flex flex-column align-items-center justify-content-center bee-without-header-height-container p-0" fluid>
+            <div
+              className="bg-img-hero d-flex flex-column align-items-center justify-content-center gradient-overlay-half-primary-v1 bee-without-header-height-container w-100"
+              style={{ backgroundImage: `url('https://static.beenest.com/images/app/misc/painted-ladies2.jpg')` }}>
```
woeltjen: This looks almost the same as the `Jumbotron` from `Home`; would it make sense to convert these into a shared component? (Or am I missing some of the differences?)

    <Beebotron>
      <h5 className="...">No listing in this area, please try a different city or region.</h5>
    </Beebotron>

kcvan: they have a different height.  i agree that theyre very similar. the main thing i dont like is that we're using bootstrap/theme classes for some places and `style` for others. 

but yes, i think we can benefit from a shared component that accepts class names and styles, and src.  will make in the next PR

src/pages/search/Search.tsx
kcvan: sounds good! i'll make a separate PR for this

src/pages/search/Search.tsx
kcvan: great idea! will be much easier to navigate

src/pages/search/Search.tsx
```diff
@@ -29,14 +32,75 @@ const Search = () => {
...
+      {({ loading, error, data }) => {
+        if (loading) {
+          return (
+            <Container tag={Fade} className={VIEWPORT_CENTER_LAYOUT}>
+              <Loading height="8rem" width="8rem" />
+            </Container>
+          );
+        }
+
+        if (error) isBadUserInputError(error) ? <SearchErrorPage /> : <ErrorPage error={error} />;
```
woeltjen: Looks like this is missing a `return`

`http://localhost:4200/work/search?locationQuery=` gives "We're sorry — something's gone wrong."

When I add a `return` statement I instead see the (really nice-looking!) "Please enter in a city or region to see results" page

src/pages/search/Search.tsx
```diff
@@ -29,14 +32,75 @@ const Search = () => {
...
+        
+        const { searchListings } = data;
+        if (searchListings.length < 1) <EmptySearchPage />;
```
woeltjen: Looks like a missing `return` here as well

`http://localhost:4200/work/search?locationQuery=The+Moon` shows a page with an empty map, but when I add a `return` here I get "No listings in this area, please try a different city or region."

(Also, might be out of scope for this issue, but should we have a CTA for "Be the first to create a listing in this area"? Based on activity from the new-hosts channel it seems like searching and then creating a listing is a pretty common use case.)

src/pages/search/Search.tsx
kcvan: 🙏

src/pages/search/Search.tsx
kcvan: would be a good idea! our bag of nice to haves is growing😂

PR #254, opened 2019-03-06
Lay out ListingInformation

woeltjen:
## Description
Layout tweaks to the Listing Information section of the listing page

## How to Test
1. Go to `/work/listings/:id` for some listing
2. Verify that layout of information in left hand column is reasonable

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. View Photos functionality
2. Booking card functionality

### DIFF
--- a/src/components/work/Avatar/Avatar.tsx
+++ b/src/components/work/Avatar/Avatar.tsx
@@ -1,11 +1,10 @@
 import * as React from 'react';
-import { Fade } from 'reactstrap';
 
-import { User } from 'networking/user';
+import { Host } from 'networking/listings';
 import LazyImage from 'shared/LazyImage';
 
 interface Props {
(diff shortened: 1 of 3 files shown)


tc:
screen shot 2019-03-05 at 4 31 59 pm

seeing an error

woeltjen:
> seeing an error

Oops! Forgot that some of those properties were only added recently. Added some conditional rendering for compatibility with older listings

tc:
lgtm; the subheaders should be `h3` or something other than `h1`

PR #256, opened 2019-03-06
Jeremy/eng 1068a host scaffolding

jerminatorhits:
## Description
Why did you write this code?
Rebranding

## How to Test
- [ ] Visit /work/host/
- [ ] Click Tabs and try various screen sizes

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


src/pages/Work.tsx
```diff
@@ -16,10 +20,14 @@ import Signup from './signup';
...
+import '../styled/customStyles.scss';
 
-import Header from 'components/work/Header';
 
-import '../styled/customStyles.scss';
+const HostWithZendesk = (props: RouterProps) => (
+  <ZendeskWebWidgetWrapper>
+    <Host {...props} />
+  </ZendeskWebWidgetWrapper>
+);
```
woeltjen: Hm, thinking it could get unwieldy to include this level of detail here (e.g. when other routes need wrappers). I like that `ZendeskWebWidgetWrapper` is not included in `Host.tsx`, that's a good separation of concerns; maybe we can introduce an intermediary component to handle this wrapping external to `Work.tsx`?

I'm picturing:

* /host/HostPage.tsx (currently Host.tsx in this PR)
* /host/Host.tsx (ZendeskWebWidgetWrapper around HostPage)
* /host/index.ts (just exports Host.tsx)

src/pages/host/Host.tsx
```diff
@@ -0,0 +1,105 @@
...
+            onClick={handleNewListingClick}>
+            Add New Listing
+          </Button>
+        </Col>
+      </Row>
+      <hr />
+      <Nav className="mb-5" tabs>
+        {[
+          {
+          to: '/work/host/bookings',
```
woeltjen: Small whitespace irregularity here

kcvan: 👀

kcvan: also, maybe can probably move this out to a `const` later?  might keep the jsx a little cleaner

kcvan:
besides small comments scaffolding looks great!

src/pages/host/Host.tsx
woeltjen: As a bonus, if we moved this to a `const`, I think we could also generate the `AuthenticatedRoute`s in the `Switch` using that same array (minor improvement. but keeps the `to` paths from falling out of sync, e.g. when we drop the `/work`)

src/pages/Work.tsx
jerminatorhits: +1 very clean architecture!

src/pages/host/Host.tsx
jerminatorhits: ahh very nice

src/pages/host/HostPage.tsx
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react';
+import { Redirect, Route, Switch, NavLink as RRNavLink } from 'react-router-dom';
+import { Listing, CREATE_LISTING, GET_HOST_LISTINGS } from 'networking/listings';
+import { Container, Fade, Button, NavItem, NavLink, Nav, Row, Col } from 'reactstrap';
+import AuthenticatedRoute from 'HOCs/AuthenticatedRoute';
+import NotFound from '../notFound';
+import { compose, graphql } from 'react-apollo';
+import { getFriendlyErrorMessage } from 'utils/validators';
+
+const HostRoutes = [
```
woeltjen: Small point, but the naming here is inconsistent; UpperCamelCase makes this look like either a class or a React component, given the conventions we use elsewhere. I think `HOST_ROUTES` would be most consistent with the way we've named other top-level constants.

jerminatorhits: kk thanks! I was confusing the use of UPPERCASE with enums, which are in a sense "constants" themselves.

PR #255, opened 2019-03-06
Show Carousel on View Photos

woeltjen:
## Description
Implement the View Photos button under the new design, this time using our existing carousel

## How to Test
1. View `/work/listings/:id` for some listing you like
2. Click "View Photos"
3. Expect photo carousel to appear full screen
4. Click X in top-right of photo carousel
5. Expect photo carousel to go away

Which devices did you test on?
- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:
Check against these principles:

### High level
Does this code need to be written?
What are the alternatives?
Will this implementation become a support issue?
How much error margin does this solution have?

### Code
* Does the code follow industry standards?
JS: https://github.com/airbnb/javascript
React: https://github.com/airbnb/javascript/tree/master/react
https://github.com/vasanthk/react-bits
Documentation headers: http://usejsdoc.org/index.html

* Is there duplicated code? Can it be refactored into a shared method?
* Is the code consistent with our project?
* Are there unit tests? Do they test the states?
* Is the person refactoring another developer's code? If possible, did the original developer approve?

### Variables/Naming:
* Would the variable type led to future edge cases?
* Are the variable naming clear? Would the value contain something other than what the name describes.

### Security
* Can this be hacked or abused by the user?

## Further Work
1. Booking Card functionality
2. And oof, this is a reminder that those left/right arrows are mispositioned; will hot fix that separately

## Learnings
After #240, I have a new appreciation for the fine art of giving up and starting over.

PR #257, opened 2019-03-06
Reposition carousel arrows

woeltjen:
## Description
Arrows on our carousel could be observed to sometimes disappear behind the image or misalign too far to the right, and to bounce around as image size changed. This positions both arrows consistently outside of the carousel photo area.

## How to Test
1. Navigate to a specific `/listings/:id`
2. Click View Photos
3. Click left and right to scroll through photos
4. Expect left/right arrows to remain in place and visible as photo dimensions change

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/styled/containers/Carousel.container.ts
+++ b/src/styled/containers/Carousel.container.ts
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 import { color, typography } from 'styled/utils';
 
 const CarouselContainerMobile = styled.div`
-  width: 320px;
+  width: calc(100vw - 80px);
 
   .carousel-image-wrapper {
     outline: 0;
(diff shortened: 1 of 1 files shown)


kcvan:
i think something might have gotten introduced in a different pr but i cant click on the view photos button to get to the carousel

woeltjen:
> i think something might have gotten introduced in a different pr but i cant click on the view photos button to get to the carousel

Are you looking at `/work/listings/:id` or just `/listings/:id`? This PR targets the latter (tho should get picked up by the former when #255 is merged)

kcvan:
ah! my bad, will checkout `/listings/:id`

kcvan:
everything looks good on desktop/tablet/mobile!  i think the only thing is that the arrows get lost on iPhone 5/SE

woeltjen:
> i think the only thing is that the arrows get lost on iPhone 5/SE

Ah, nice catch! Marking WIP to fix that

PR #258, opened 2019-03-06
add Footer to work routes, make mobile listing page footer disappear

jerminatorhits:
## Description
Why did you write this code?
Move Footer to pages/work so it renders automatically

## How to Test
- [ ] Visit /work/home, /work/account, /work/trips, /work/search and see Footer
- [ ] Visit /work/listings/:id and see Footer on desktop but no Footer on tablet/mobile

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/components/work/Footer/Footer.tsx
+++ b/src/components/work/Footer/Footer.tsx
@@ -69,9 +69,10 @@ const socialData = [
   },
 ];
 
-const Footer = () => (
+export const Footer = () => (
   <Switch>
     <Route path="/work/account" component={DetailedFooter} />
+    <Route exact path="/work/listings/:id" component={NoopComponent} />
     <Route exact path="/work" component={DetailedFooter} />
     <Route exact path="/work/about" component={DetailedFooter} />
     <Route exact path="/work/login" component={NoopComponent} />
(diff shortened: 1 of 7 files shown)


src/pages/listing/Listing.tsx
```diff
@@ -20,7 +18,7 @@ const Listing = ({ match }: RouterProps) => (
         <ListingPage {...data.listing} />
       }
     </Query>
-    <Footer />
+    <DetailedFooter className="d-none d-lg-block" />
```
kcvan: actually now that `<DetailedFooter>` takes a classname, i think you can just export the default `<Footer>` and use react router to handle the all the footers again

src/components/work/Footer/Footer.tsx
```diff
@@ -69,9 +69,10 @@ const socialData = [
   },
 ];
 
-const Footer = () => (
+export const Footer = () => (
   <Switch>
     <Route path="/work/account" component={DetailedFooter} />
+    <Route exact path="/work/listings/:id" component={NoopComponent} />
```
kcvan: we can change `component=` to `render=` and pass down className as a prop for the component

src/pages/listing/Listing.tsx
jerminatorhits: 🙏

src/components/work/Footer/Footer.tsx
jerminatorhits: ty

kcvan:
LGTM!

tc:
@jerminatorhits  conflict

PR #261, opened 2019-03-06
Add state to BookingCard

woeltjen:
## Description
Adding functionality into BookingCard for the work theme. Breaking this up into smaller chunks to keep the diffs simple.

## How to Test
1. Visit `/work/listings/:id` for some listing you like
2. Verify that form in BookingCard is modifiable

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. Block out reserved dates
2. Define date range for `isOutsideDateRange`
3. Dynamic pricing
4. Initialize from query parameters
5. Create bookings on Request to Book

## Learnings
`DateRangePicker` from `react-dates` is a controlled component; you need to give it back the `focusedInput` from `onFocusChange` in order to display the pickers.

### DIFF
--- a/src/pages/listing/BookingCard.tsx
+++ b/src/pages/listing/BookingCard.tsx
@@ -19,6 +19,7 @@ const BookingCard = ({
   const [focusedInput, setFocusedInput] = React.useState<'startDate' | 'endDate' | null>(null);
   const [startDate, setStartDate] = React.useState<moment.Moment | null>(null);
   const [endDate, setEndDate] = React.useState<moment.Moment | null>(null);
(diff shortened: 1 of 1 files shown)


kcvan:
who needs state when we have HOOOOOKS

woeltjen:
> card looks amazing, vic! there's no functionality for dynamic pricing/mobile/tablet yet, right?

Yep, that's correct. (I forgot about BookingBar, I'll add some notes under future work in the next PR; thanks!)

PR #268, opened 2019-03-06
Request to Book from BookingCard

woeltjen:
## Description
Continued migration of Listing page to the new design: Implement "Request to Book" in the BookingCard

## How to Test
1. Navigate to a listing
2. Expect "Request to Book" to be disabled without dates
3. Choose some dates
4. Expect "Request to Book" to enable
5. Click "Request to Book"
6. Expect to be taken to `/bookings/:id` for the new booking

Note that `/bookings/:id` does not immediately load; think this is related to going from `/work` to `/`, but not sure. In any event, expect that to be addressed by migration of booking pages to the new design

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
BookingBar!

### DIFF
--- a/src/pages/listing/BookingCard.tsx
+++ b/src/pages/listing/BookingCard.tsx
@@ -1,13 +1,16 @@
 import * as React from 'react';
 import { Button, Card, Input, Row } from 'reactstrap';
 import moment from 'moment';
-import { Query } from 'react-apollo';
+import { compose, graphql, Query } from 'react-apollo';
+import { withRouter } from 'react-router';
 
 import DateRangePicker from 'components/work/DateRangePicker';
 import { guestsSelectboxOptions } from 'components/work/SearchBar/searchBar.config';
 import Loading from 'shared/loading/Loading';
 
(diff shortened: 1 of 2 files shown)


src/pages/listing/BookingCard.tsx
```diff
@@ -110,6 +122,51 @@ const BookingCard = ({
...
+  async function startBooking() {
+    setBooking(true);
+    try {
+      const { data } = await createBooking({
+        checkInDate: String(startDate),
+        checkOutDate: String(endDate),
+        listingId: id,
+        numberOfGuests: numberOfGuests
+      });
+      history.push(`/bookings/${data.createBooking.id}`);
```
kcvan: i think for now we have to do `window.location.href` because history.push uses react router from what i remember

woeltjen: Ah, think that explains why I needed to reload to see the booking change. Thanks!

PR #274, opened 2019-03-07
Add Transit Time filter

woeltjen:
## Description
The ability to search for hotels by their relative proximity to a destination will add value for business travelers, who frequently travel with a specific conference location or work site as their destination.

## How to Test
1. Start at `/work`
2. Search for a city or region, e.g. "Los Angeles County, CA, USA"
3. Click "Transit Time"
4. Enter a specific location, e.g. "Churros near Sleeping Beauty Castle"
5. Expect to see search results near that location near the top

Note that UX for Transit Time changes is a little inconsistent; you'll see the loading indicator the first time you make a specific search, but if you repeat that search the results will just reorder (I assume this is caching from GraphQL). Think this is a nuance we can carry for now.

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. Show distinct pin at destination
2. Retain place selection in dropdown
3. Display driving directions
4. Moar filters

## Learnings
Google Places supports a [lot of place types](https://developers.google.com/places/supported_types); in lieu of trying to prune those down preemptively, I'm just allowing everything for now.

### DIFF
--- a/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
+++ b/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
@@ -10,6 +10,7 @@ interface Props {
   onPlaceChange(place: google.maps.places.PlaceResult): void;
   inputRef: React.RefObject<HTMLInputElement>;
   children: React.ReactNode;
+  types?: string[];
 }
 
 
@@ -25,9 +26,10 @@ class GoogleAutoComplete extends React.Component<Props, any> {
     if (!this.props.inputRef.current) return;
     if (!window.google) return;
 
+    const types = this.props.types || ['(regions)'];
     this.autocomplete = new google.maps.places.Autocomplete(
       this.props.inputRef.current,
-      {"types": ["(regions)"]}
+      { types }
     );
     if (this.autocomplete.setFields) this.autocomplete.setFields(['geometry', 'name']);
     this.autocomplete.addListener('place_changed', this.handlePlaceChanged);
--- a/src/networking/listings.ts
+++ b/src/networking/listings.ts
@@ -62,9 +62,10 @@ export interface ListingSearchCriteria {
   checkInDate?: string;
   checkOutDate?: string;
   numberOfGuests?: number;
-  locationQuery: string;
+  locationQuery?: string;
   bounds?: LatLngBounds;
   coordinates?: LatLng;
+  near?: LatLng;
 }
 
 export interface ListingShort {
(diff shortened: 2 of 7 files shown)


woeltjen:
Removed WIP status; think this is a reasonable increment to review and merge. See notes in Further Work and How to Test for reference on completeness of this implementation in its current state.

kcvan:
Looks good, vic!

some things that i noticed (might not be needed to implement later):
1) is the map supposed to center around the location that you're searching?
2) `Transit Time` makes me think that im going to get the times between the location and hotels in the area, should we change it to something like `Close To` or something similar?
3) might be out of the scope of this pr, but is it possible to throw a different colored pin when searching for a specific location so we can tell where it is on the map?

woeltjen:
> 1. is the map supposed to center around the location that you're searching?

Not sure! My thinking is to punt until Driving Directions are implemented. The map will jump around as the user selects listings, at that point, which might give us a better feel for what the initial positioning should be.

> 2. Transit Time makes me think that im going to get the times between the location and hotels in the area, should we change it to something like Close To or something similar?

I agree; I started off with "Distance From", but decided to stick with the design. I'll change text to "Close To..." and loop in Hiroo as to why. (When Driving Directions are implemented, the "Transit Time" name might make more sense.)

> 3. might be out of the scope of this pr, but is it possible to throw a different colored pin when searching for a specific location so we can tell where it is on the map?

Good point! I'll add another line item for that under Further Work and handle in a subsequent PR

kcvan:
`
Not sure! My thinking is to punt until Driving Directions are implemented. The map will jump around as the user selects listings, at that point, which might give us a better feel for what the initial positioning should be.
`

I guess my initial thought was, if we're entering in a location that we want to be near in the filter, centering that location and seeing available listings around that would be ideal.  but im not too sure either, since we're doing something pretty new.

@tommy-bee @jerminatorhits  thoughts?

kcvan:
also out of scope for this pr, but hovering over a listing and having the locator highlight and card pop up on the map would be a really great to have🙈

kcvan:
I think as it is right now its pretty great for this pr!  my only gripe is that the entire pages loads instead of the listing cards and the map.  this can probably be approved upon later by moving some things around.

LGTM

woeltjen:
> my only gripe is that the entire pages loads instead of the listing cards and the map. this can probably be approved upon later by moving some things around.

Heh, yes, that has been bugging me too. I'm shifting to a controlled component model for the next PR ("Retain place selection in dropdown") and will see if that helps, but otherwise, yeah, think we can carry it for a bit.

PR #276, opened 2019-03-07
Jeremy/host-portal-rebrand

jerminatorhits:
## Description
Why did you write this code?
Rebrand host/bookings

## How to Test
- [ ] Visit host/
- [ ] Verify functionality of pages/tabs
- [ ] If logged in as a user with a listing, the header button should read "Host Profile" and take you to host portal '/host'
- [ ] If logged in as a user without a listing or not logged in, it should read "Become a Host" and take you to signup pages
- [ ] Test stripe routes on host/payment

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


src/legacy/routes/Host/HostBookingCard/HostBookingCard.tsx
```diff
@@ -14,6 +11,9 @@ import {
 } from 'utils/bookingsDisplayStatus';
 import { currencyToDisplay } from 'utils/currencyToDisplay';
 import { formatDateRange } from 'utils/formatDate';
+import { Card, Col, Row } from 'reactstrap';
+import Loading from '../../../shared/loading/Loading';
```
kcvan: can we use the `/legacy/shared/etc` import here? will be easer to change all occurrences later

src/legacy/work/Header/Header.tsx
```diff
@@ -68,33 +71,58 @@ const DetailedHeader = () => {
...
+                                  {item.header}
+                                </NavLink>
+                              </NavItem>
+                            ))}
+                          </>
+                        );
+                      }}
+                    </Query>
+                  )
+                }
```
jerminatorhits: All this just to dynamically render the Host Portal button name on the header O_O

src/legacy/routes/Host/HostBookingCard/HostBookingCard.tsx
jerminatorhits: 👍

src/legacy/work/Header/Header.tsx
kcvan: 🙈

src/legacy/work/Header/Header.tsx
```diff
@@ -68,33 +71,58 @@ const DetailedHeader = () => {
...
+                        return (
+                          <>
+                            {beeUser.listingCount > 0
+                              ? <Link to={HOST_PORTAL_LINK} className="mb-4 mb-md-0 mr-md-4 w-100 w-md-auto btn btn-outline-primary">
+                                  Host Profile
+                                </Link>
+                              : <Link to={HOST_INTEREST_LINK} className="mb-4 mb-md-0 mr-md-4 w-100 w-md-auto btn btn-outline-primary">
+                                  Become a Host
+                                </Link>
+                            }
```
woeltjen: Looks like there's a small opportunity for deduplication here with a

    const isHost = beeUser.listingCount > 0;

And then a

    <Link to={isHost ? HOST_PORTAL_LINK : HOST_INTEREST_LINK} className="letter-hyphen-number-party">
      {isHost ? 'Host Profile' : 'Become a Host'}
    </Link>

src/legacy/work/Header/Header.tsx
woeltjen: Worth moving some of this into its own component? The nesting gets pretty deep, but 142 LOC still seems reasonable for the Header as a whole, so I'm neutral

src/legacy/work/Header/Header.tsx
jerminatorhits: 🙏

src/legacy/work/Header/Header.tsx
jerminatorhits: Hmmm nesting does get pretty deep. I don't anticipate the Header changing very much from this point so I think it'd be good to keep logic local for now.

PR #280, opened 2019-03-07
Remove Fade from GoogleAutoComplete

woeltjen:
## Description
Noticed in the context of #274 that the GoogleAutoComplete input was fading in. The delay was a bit distracting during testing, which reminded me of @tc's [earlier comment](https://github.com/thebeetoken/beenest-web/pull/234#discussion_r262193706) about slowing the user down.

This removes `Fade` from GoogleAutoComplete. You can see an example of the change at `/work` (where, with Fade, the location input fades in *just a bit* slower than the rest of the UI).

Thinking that it makes sense to utilize `Fade` at the level of pages and cards and so on, instead of fine-grained components like form controls; in the latter case, it seems too hard to know if the animation will be appropriate or not.

## How to Test
Regression-test search from the front page (or any other GoogleAutoComplete usages you can think of)

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
+++ b/src/components/shared/GoogleAutoComplete/GoogleAutoComplete.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { Fade } from 'reactstrap';
 import { compose, withProps } from 'recompose';
 import { withScriptjs } from 'react-google-maps';
 
@@ -38,11 +37,7 @@ class GoogleAutoComplete extends React.Component<Props, any> {
   }
 
   render() {
-    return (
-      <Fade>
-        {this.props.children}
-      </Fade>
-    );
+    return this.props.children;
   }
 }
 


tc:
+1 cc @kcvan

kcvan:
i think i added the fade because the autcomplete actually has a bit of delay when loading, so it just flahes in.  if you hard refresh you can see it coming in.  im not against removing the fade, just wanted to make it a little less choppy when it loads for the first time.

woeltjen:
> i think i added the fade because the autcomplete actually has a bit of delay when loading, so it just flahes in. if you hard refresh you can see it coming in. im not against removing the fade, just wanted to make it a little less choppy when it loads for the first time.

Ah, indeed! It's subtle but it's there. (I had to put two windows side-to-side, one with an input and one with a GoogleAutoComplete-wrapped input, before I convinced myself I could see it, but once I did I was able to tell which was GoogleAutoComplete just from that "pop" on load.)

It looks like it's `withScriptjs` that is introducing the delay. Even after first load, it seems to add a one-frame delay any time a `withScriptjs`-wrapped component is shown. We could move the Google Maps dependency to a script tag in the `index.html` template; it's a bit more bloat to the initial load, but we are using Google Maps pretty ubiquitously (there's not a lot you can do on the site *without* loading it).

kcvan:
i think the delay is small enough to ignore it for now.  i think the natural input delay + the fade delay made it seem much slower than it actually is so i think this is a good change 👍

PR #284, opened 2019-03-08
deploy

tc:
big deploy for business rebrand.

smoke test on staging.beenest.io


woeltjen:
Tried to make a booking, everything went smoothly up until I got to the receipt page, which just showed with the blue header and a white page below. Reloading showed the receipt page as expected

woeltjen:
Hm, can't reproduce the above error now. Not sure what the issue was, but (a) the workaround was reasonable, just reload, and (b) it's not regularly reproducible.

So, approving.

PR #289, opened 2019-03-08
Show marker at 'Close To...' location

woeltjen:
## Description
Continuing with search filters: Show a marker at the user's chosen 'Close To...' location

## How to Test
1. Search
2. Click 'Close To...'
3. Select some place on your map
4. Verify that labelled marker shows up at that place

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
1. Driving directions
2. Centering on pin
3. More filters

## Learnings
Looked at the "centering on pin" part, think that needs to be its own issue; correct behavior is non-obvious, as other variables including the search query's bounds and the locations of listings themselves need to be considered

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -62,6 +62,15 @@ class GoogleMapsWithMarkers extends React.Component<Props, State> {
         ref={this.handleMapMounted}
       >
         {near && <Marker
+          icon={{
+            fillColor: '#AADBFF',
+            fillOpacity: 1,
+            labelOrigin: { x: 0, y: -5 },
+            scale: 4,
+            strokeColor: '#1164FF',
+            path: google.maps.SymbolPath.CIRCLE,
+            strokeWeight: 2
+          }}
           label={near.name}
           position={near.geometry.location}
           title={near.name}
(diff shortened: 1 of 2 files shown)


kcvan:
pin shows! is it possible to make the pin look like the red ones...but blue?  the pin and text were a little hard to find.  wondering if centering the map around the pin would be better ui?

@tommy-bee @jerminatorhits

woeltjen:
Changing the color is unfortunately [non-trivial](https://stackoverflow.com/questions/2472957/how-can-i-change-the-color-of-a-google-maps-marker); we get to replace the whole icon or use their default, basically. Having our own icon image set would be nice.

Agree on centering; see notes under "Learnings".

kcvan:
> Agree on centering; see notes under "Learnings".

ah! sorry i missed that. 👍

kcvan:
>  Having our own icon image set would be nice.

time to get a bee svg

kcvan:
https://iconmonstr.com/location-1-svg/ maybe?

woeltjen:
> https://iconmonstr.com/location-1-svg/ maybe?

Ooh, I like! Filed #295 to look at icons separately; there were a couple in the same set as that one that looked promising to me

PR #290, opened 2019-03-08
Show driving directions

woeltjen:
## Description
Business travelers care about being close to specific locations (conference halls, work sites, and so on) so let's help them out!

Note that this has been branched from #289 and should be reviewed/merged after that has been reviewed/merged.

## How to Test
1. Search (e.g. for "San Francisco")
2. Specify a "Close To..." filter (e.g. "Moscone Center")
3. Click on listings on the map
4. Expect to see the route to your destination appear

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Playing with this, there are a variety of UX improvements I might want to see. Think we should play around with this, discuss, and itemize any non-blockers with Jira tickets.

## Learnings
I'm just generally impressed with the API design of Google Maps. Easy to do hard things but not hard to do easy things!


src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
```diff
@@ -25,12 +26,18 @@ interface State {
...
+  directionsService = new google.maps.DirectionsService();
+  directionsDisplay = new google.maps.DirectionsRenderer();
+
   state: State = {}
 
   handleMapMounted = (map: GoogleMap) => {
     if (!map) {
       return;
     }
+    const googleMap: google.maps.Map = map.context.__SECRET_MAP_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
```
woeltjen: This is stupid. `react-google-maps` doesn't expose the `google.maps.Map`, but the `google.maps.DirectionsRenderer` needs it. `react-google-maps` is also unmaintained at this point. There are other React Google Maps libraries out there, but that makes the scope grow.

Here are the options I see:

1. Use the hack above (at the risk of *being fired*) and add an annotation to silence TypeScript's complaining.
2. Switch to another React Google Maps library (which is a slightly larger refactor which would block us here)
3. Use the Google Maps API directly from here (and elsewhere, progressively; think we started this with Autocomplete?)

The engineer in me is drawn toward Option 1, as it's the simplest progressive enhancement which does not preclude any other options.

The developer in me is drawn toward Option 2, because we're going to need to do that anyway and intermediary solutions are just overhead.

The programmer in me is drawn toward Option 3, because a React API for Google Maps doesn't do anything more than the JS API for Google Maps, so why bother?

So, soliciting advice!

tc: It seems all of our google maps code is abstracted behind the `GoogleMaps` and `GoogleAutoComplete` components so taking on `3` isn't a bad choice especially if the react api is just a wrapper and (unmaintained).

kcvan:
DRIVING DIRECTION FEATURE IS SO COOL

src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
kcvan: firstly, i second, the third option also

woeltjen: After all that, it turns out `react-google-maps` has a [`DirectionsRenderer` component](https://github.com/tomchentw/react-google-maps/blob/ca5c5c6f5346fb3ee0037893fbca488a408c9938/src/components/DirectionsRenderer.md), so there's not a need to try to get at the `google.maps.Map` directly.

Oops.

jerminatorhits: lolol. react-google-maps has been so hassle as app grows ><.

woeltjen:
Wow, somehow this ended up being a very simple diff!

Removed WIP status, think this is ready.

PR #292, opened 2019-03-08
Select search results on hover

woeltjen:
## Description
Show info window and/or driving directions for search results when user hovers over them

## How to Test
1. Search for listings in a city
2. Hover over listing results on the left
3. Expect info windows to appear on the right
4. Enter a "Close To..." location
5. Hover again
6. Expect to see driving directions appear on the right

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security


kcvan:
feature loooks awesomeeeee.  i know it's still wip but some things i noticed:
1) hovering around the listing card causes flashes on the listing map card.  should we just make the entire card clickable?
2) hovering on a listing moves the map around, should we keep the map fixed as long as the listing is in view of the map?

woeltjen:
> 1. hovering around the listing card causes flashes on the listing map card. should we just make the entire card clickable?

Currently clicking navigates to the listing. If it selects, what navigates? Second click is an option (and would be more touchscreen-friendly) but I felt hesitant to introduce another user action between searching and booking.

Do you think debouncing would mitigate the flashing sufficiently?

> 2. hovering on a listing moves the map around, should we keep the map fixed as long as the listing is in view of the map?

Right now I'm mostly letting the map do its own thing. In this case, it's zooming in on driving directions and then making room for the info window, which definitely results in more movement than needed.

I'd be inclined to call this a future UX improvement, assuming the current behavior seems acceptable to roll out to users. (Which may be a false assumption.) There are a variety of other UX problems (e.g. the way the info window hides the driving directions, some times) that should be addressed, but don't need to block. I'll schedule a meeting next week to review the feature and itemize/prioritize future work.

kcvan:
> Do you think debouncing would mitigate the flashing sufficiently?
is it flashing because the hover action is being called on separate components in the card?

> Right now I'm mostly letting the map do its own thing. In this case, it's zooming in on driving directions and then making room for the info window, which definitely results in more movement than needed.
>
>I'd be inclined to call this a future UX improvement, assuming the current behavior seems acceptable to roll out to users.

i think the jumping is a little distracting but i would definitely say it can be a later ux improvement

jerminatorhits:
debounce will be a good future feature to implement. When I hover of many listings at a time, sometimes the line doesn't show.

The flashing does become a bit much on the eyes as user hovers over different listings. My preference would be to scrap the flashing for the cards on the map. We can discuss the direction of the features next week

woeltjen:
Added some debounce (thanks @jerminatorhits for finding that `useDebounce` snippet; refactored it into its own util). Finding a good duration was a bit tricky; too high introduces a noticeable delay on hover feedback, too low gets flashy. 125ms seems to work okay (iirc humans tend to perceive feedback as "delayed" after about 150ms, but that's a squishy metric)

Removing WIP status as this seems reasonable for now (insofar as I don't see blockers)

> is it flashing because the hover action is being called on separate components in the card?

I don't *think* so, because `onMouseOver` is only on the `Link` that wraps the whole card. But I might be getting my `mouseover` semantics wrong. Should this case be `enter` and `leave`?

woeltjen:
Yep, `over`/`out` were the problem (or part of it); switched to `enter`/`leave`

kcvan:
Woo! the hover feels much smoother now that the flashing is gone!

i think the only strange interaction i had was when combining the `Close To...` with the hover.  One or two times the destination will be almost off the map when the listing is hovered on, but i havent found another listing that does this besides this one:
Screen Shot 2019-03-11 at 11 57 55 AM

so i dont think it's a huge deal.  great features, vic! pumping them our like crazy 🙏

PR #295, opened 2019-03-08
Change icons for map search

woeltjen:
## Description
Follow up on discussion from #289: Improve icons for hotels and destinations

## How to Test
1. Search for some hotels
2. Enter a "Close To..." location
3. Do the markers on the map make sense?

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security


woeltjen:
image

Fond of the ★ marker; I recall that in the UI prototype we had some trouble distinguishing hotels from the other destinations, trying out the combination of home-vs-star to see if that helps.

tc:
star as destination works. 
home as home/hotel works.

where's the driving path from DG to coit tower in the map?

kcvan:
> where's the driving path from DG to coit tower in the map?

*running path😂

kcvan:
looks great, vic!

src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
```diff
@@ -11,12 +11,16 @@ import { LatLngBounds, ListingShort } from 'networking/listings';
 
 import GoogleMapsWithMarkersContainer from './GoogleMapsWithMarkers.container';
 
+const hotelMarker = require('assets/images/iconmonstr-location-12-32.png');
```
kcvan: i think we can upload these to s3 and use it there

tc: either way, the icons are still cached via cloudfront.

PR #300, opened 2019-03-11
share buttons on listing gallery

kcvan:
## Description
Why did you write this code?
add share buttons on listing gallery
## How to Test
- [ ] Visit any listing
- [ ] Click the share button on top right
- [ ] Verify they all work

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/legacy/work/ShareModal/ShareModal.tsx
+++ b/src/legacy/work/ShareModal/ShareModal.tsx
@@ -65,7 +65,7 @@ const ShareModal = (props: Props) => {
 
   function toggleModal() {
     toggle();
-    setAlert(!isAlertOpen);
+    setAlert(false);
   }
 
   function toggleAlert() {
--- a/src/legacy/work/ShareModal/ShareModal.tsx
+++ b/src/legacy/work/ShareModal/ShareModal.tsx
@@ -12,8 +12,8 @@ const ShareModal = (props: Props) => {
   const [isAlertOpen, setAlert] = React.useState<boolean>(false);
   const { isOpen, toggle, url, title } = props;
   return (
-    <Modal isOpen={isOpen} toggle={() => toggle()}>
-      <ModalHeader className="text-primary" toggle={() => toggle()}>Share Listing</ModalHeader>
+    <Modal isOpen={isOpen} toggle={() => toggleModal()}>
+      <ModalHeader className="text-primary" toggle={() => toggleModal()}>Share Listing</ModalHeader>
       <ModalBody>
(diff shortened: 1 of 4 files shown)


kcvan:
*note

i had to use 

```
const listingLink = `https://beenest.com${location.pathname}`;
```
instead of `${BEENEST_HOST}` because facebook didnt like linking from a localhost

src/pages/listing/ListingGallery.tsx
```diff
@@ -1,20 +1,89 @@
...
+  const [isAlertOpen, setAlert] = React.useState<boolean>(false);
+
+  const { listingPicUrl, location, photos, title } = props;
+  const listingLink = `https://beenest.com${location.pathname}`;
   return <Row className="w-100 height-60vh px-0 mx-0 position-relative">
     <ImageGrid images={[listingPicUrl, ...photos]} />
+    <Button className="position-absolute top-0 right-0 m-4" color="primary" onClick={() => toggleModal()} type="button">
+      <span className="far fa-share-square"></span>
+    </Button>
+    <Modal isOpen={isModalOpen} toggle={() => toggleModal()}>
```
woeltjen: This looks good! Can we move this out to its own component? Picturing usage like:

    <ShareModal isOpen={...} toggle={...} url={listingLink} title={title} />

With this component inline, most of the lines of code in `ListingGallery` end up being about the sharing button, which is unexpected.

kcvan: makes sense! will move it out to it's own component

src/legacy/work/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+            Copied to clipboard.
+          </Alert>
+        </Row>
+      </ModalBody>
+    </Modal>
+  );
+
+  function toggleModal() {
+    toggle();
+    setAlert(!isAlertOpen);
```
jerminatorhits: When I open and close the modal, I think the alert is opening and closing as well. Maybe `setAlert(false)` for this line?

kcvan: thanks! copypasta is not always the most delicious pasta

PR #303, opened 2019-03-11
Show an alert when place selection fails

woeltjen:
https://github.com/thebeetoken/beenest-web/pull/301#issuecomment-471770175

## Description
Proposed followup to #301: Show an alert when a user's place selection is rejected because they didn't pick from the list.

## How to Test
1. Conduct a search
2. Start typing in the "Close To..." text input
3. Click enter
4. Expect an alert to appear instructing you to choose from the autocomplete results

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/pages/search/filters/TransitTime.tsx
+++ b/src/pages/search/filters/TransitTime.tsx
@@ -12,8 +12,12 @@ const TransitTime = ({ place, onPlaceChange }: Props) => {
   const inputRef: React.RefObject<HTMLInputElement | null> = React.createRef();
 
   const handlePlace = (place: google.maps.places.PlaceResult) => {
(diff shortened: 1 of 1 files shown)


kcvan:
works! would it be possible to use the `<Alert />` component and just have it show up above the search bar if the user presses enter?

woeltjen:
> works! would it be possible to use the `<Alert />` component and just have it show up above the search bar if the user presses enter?

Yep, that looks much nicer

image

PR #304, opened 2019-03-11
Make search filter full width

woeltjen:
## Description
To reduce likelihood of place names being cut off, make the dropdown for a search filter the full width of its parents.

## How to Test
1. Search
2. Click "Close To..."
3. Expect dropdown to use full width (sans margin/padding) of its containing element

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security


woeltjen:
image

PR #305, opened 2019-03-11
Add travel mode selection

woeltjen:
## Description
This adds the ability to choose travel mode (from driving, transit, walking, and bicycling) when selecting a "Close To..." location

## How to Test
1. Make a search
2. Enter a "Close To..." location
3. Hover over some listings to view directions
4. Use "Close To..." drop down to change travel mode (e.g. to "walking")
5. Hover over more listings
6. Expect directions to change to fit the specified travel mode

Note that sometimes the visual difference between (e.g.) driving directions and walking can be subtle.

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
The full-page-reload which occurs on some destination changes becomes a bit more painful with this change, as it prevents the user from directly proceeding to selecting travel mode

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -59,7 +59,11 @@ class GoogleMapsWithMarkers extends React.Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props) {
-    if (['near', 'selectedListing', 'travelMode'].every(key => prevProps[key] === this.props[key])) {
+    if (
+      prevProps.near === this.props.near &&
+      prevProps.selectedListing === this.props.selectedListing &&
+      prevProps.travelMode === this.props.travelMode
+    ) {
       return;
     }
     const { near, selectedListing, travelMode } = this.props;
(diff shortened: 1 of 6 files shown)


woeltjen:
Marking WIP while grumble grumble TypeScript

kcvan:
i think the hovering of the card works really well!  

a ux experience i felt was that it does it a little confusing when i click on a listing on the map, then try and change the filter, the map doesn't update with new directions.  it also gets a little confusing when we have a listing clicked on the map and you move your mouse over to the filter and accidentally hover over another listing

woeltjen:
> a ux experience i felt was that it does it a little confusing when i click on a listing on the map, then try and change the filter, the map doesn't update with new directions.

Ah, good catch, there was a check on `GoogleMapsWithMarkers` that was swallowing that change. Fixed; switching among driving options can be a nice use case!

> it also gets a little confusing when we have a listing clicked on the map and you move your mouse over to the filter and accidentally hover over another listing

Hm, that hurts the above use case. I've itemized that in the spreadsheet from earlier. Not sure what the fix would be; maybe require user to remain hovered for 150ms before changing selection?

kcvan:
> Hm, that hurts the above use case. I've itemized that in the spreadsheet from earlier. Not sure what the fix would be; maybe require user to remain hovered for 150ms before changing selection?

this sounds like a solid fix!

kcvan:
> Ah, good catch, there was a check on GoogleMapsWithMarkers that was swallowing that change. Fixed; switching among driving options can be a nice use case!

works great!

PR #308, opened 2019-03-12
Show prices in markers

woeltjen:
## Description
To improve utility of the map when we search, show prices instead of generic hotel markers.

## How to Test
1. Search for some listings
2. Expect to see prices on the map instead of hotels
3. Regression-test related functionality (e.g. clicking on listings in the map)

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -114,16 +114,14 @@ class GoogleMapsWithMarkers extends React.Component<Props, State> {
             key={listing.id}
             position={{ lat: listing.lat, lng: listing.lng }}
             mapPaneName={OverlayView.OVERLAY_MOUSE_TARGET}
-            getPixelPositionOffset={(w, h) => ({ x: -w / 2, y: -h })}
           >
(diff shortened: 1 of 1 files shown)


kcvan:
prices!

noticed something a little strange when messing around.  
1) search `Coit Tower`
2) click on listing on map
3) click the close button on the listing card
4) click on another listing on the map

when i do this, the listing i just clicked shows up and then goes away.  then the map pops up a card of something that was near that location and will continue doing this until you hover over a card to 'reset' it
Screen Shot 2019-03-12 at 1 02 25 PM

woeltjen:
@kcvan Nice find, thanks for the steps to reproduce. I think what is happening there is that the built-in Info Windows (e.g. "The Ritz-Carlton, San Francisco") are conflicting with the custom, listing-specific Info Window we want to show for selected listings.

I disabled clicking on map-provided icons and could no longer reproduce thereafter. We lose a bit of functionality this way (ability to view addresses for Google-provided map markers) but I don't know of any significant use cases which depend on that functionality, so I think that's fine.

PR #310, opened 2019-03-12
Add Home Type filter

woeltjen:
## Description
Adds the "Home Type" filter from updated Beenest design

## How to Test
1. Search for some listings (e.g. in San Francisco, where we have many kinds)
2. Click "Home Type"
3. Choose a different home type
4. Expect listings to filter down to that home type
5. Regression test Destination filters

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
This exposes some UX issues that we may want to solve now, or carry:

1. Approach to dropdown width from #304 was broken (it depended on the dropdown button being in a full-width column, but that space is needed for more buttons). Solution here (explicit widths for dropdowns in rems) is not mobile friendly.
2. When a Home Type filter results in listings, the "No listings were found..." page displays, losing the context of the original search.

### DIFF
--- a/src/pages/search/SearchForm.tsx
+++ b/src/pages/search/SearchForm.tsx
@@ -15,13 +15,13 @@ const SearchForm = ({ filter, onFilterChange }: Props) => {
   return <Container>
     <Row>
       <ButtonGroup>
-        <SearchFilter label="Home Type" menuClassName="bee-search-menu-narrow">
+        <SearchFilter label="Home Type" width="9rem">
           <HomeType
             homeType={filter.homeType}
             onChange={homeType => console.log(homeType)}
           />
         </SearchFilter>
-        <SearchFilter label="Destination" menuClassName="bee-search-menu-wide">
+        <SearchFilter label="Destination" width="24rem">
           <TransitTime
             place={filter.near}
             travelMode={filter.travelMode}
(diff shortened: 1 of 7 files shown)


src/pages/search/SearchForm.tsx
```diff
@@ -13,36 +14,24 @@ interface Props {
...
-      <Col>
-        <SearchFilter label="Add Destination">
+      <ButtonGroup>
+        <SearchFilter label="Home Type" width="9rem">
+          <HomeType
+            homeType={filter.homeType}
+            onChange={homeType => onFilterChange({ ...filter, homeType })}
+          />
+        </SearchFilter>
+        <SearchFilter label="Destination" width="24rem">
```
woeltjen: "Add Destination" looked inconsistent next to "Home Type" (verb versus noun); with multiple buttons, I think the context is clear enough that just "Destination" is fine here, but want to call that out for review

kcvan: i think now that we have the home type filter next to it 'Destination' seems good now!

woeltjen:
> works! can we add `noGutters` to the row that the buttons are in?

Ah, thanks, I knew there was something off about the alignment there. Added!

PR #313, opened 2019-03-13
Check isActive for availability from BookingCard

woeltjen:
## Description
Goes with thebeetoken/beenest-backend#809: When a listing is marked as `inActive`, disallow clicking on Request to Book. This is used to make listings unavailable when no pricing is available from the Agoda API

## How to Test
1. Navigate to an Agoda listing with booked dates; I use `/listings/agoda_7005`
    1. Make sure *not* to enter dates during search
2. Enter dates into BookingCard (I use 3/23-3/25 to make the above unavailable)
3. Verify that Request to Book button is disabled
4. Find some dates with availability (4/25-4/27 with the above)
5. Verify that Request to Book button is enabled

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
There's some additional user feedback we could use to explain *why* booking is unavailable in this circumstance

### DIFF
--- a/src/networking/listings.ts
+++ b/src/networking/listings.ts
@@ -369,6 +369,7 @@ export const GET_PUBLIC_LISTING = gql`
         displayName
         profilePicUrl(width: $width, height: $height)
       }
+      isActive
       ...ListingDetails
     }
   }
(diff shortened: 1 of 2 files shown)


jerminatorhits:
tested and works as described

I think users might be confused on why they can't book and may consider this a "bug" and complain to Ops. Perhaps we can somehow render a msg saying "listing is unavailable for these dates" or something

woeltjen:
Added tooltip and a price placeholderwhen listings are unavailable:

image

PR #316, opened 2019-03-13
Catch errors loading Google Maps API

woeltjen:
## Description
We load the Google Maps JS API at run-time from another origin, so sometimes it is undefined. This adds a `componentDidCatch` and falls back to showing an error message when this occurs (same strategy is used in the GoogleMaps component, which we display on listing pages.)

May also prevent some of the map-related errors we see on Sentry; less certain of that, as I haven't been able to reproduce those.

## How to Test
1. Replace all references to `https://maps.googleapis.com/maps/api/js` in the code base to some broken URL
2. Try to search
3. Expect to see an error message in place of the map

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
In the old design, we hid the map and made room for listings when this happened. Not sure if it's worth reinvesting effort in that, since this is an off-nominal case and the map is more central to the redesign

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { withRouter } from 'react-router-dom';
-import { compose, withProps } from 'recompose';
+import { branch, compose, lifecycle, renderComponent, withProps } from 'recompose';
 import { DirectionsRenderer, InfoWindow, Marker, GoogleMap, OverlayView, withGoogleMap, withScriptjs } from 'react-google-maps';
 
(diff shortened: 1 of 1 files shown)


kcvan:
error shows!

src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
```diff
@@ -155,6 +155,14 @@ export default withRouter(
         />
       ),
     })),
+    lifecycle({
+      // @ts-ignore
```
kcvan: 😂

PR #317, opened 2019-03-13
Add banner to app

kcvan:
## Description
Why did you write this code?
add email and phone verification banner to app
## How to Test
- [ ] Create a new account
- [ ] Banner should show up on mobile/tablet/desktop
- [ ] Banner should be closable by clicking the `X` button
- [ ] Banner should lead you to account verification page
- [ ] Verifying your phone number and email, then refreshing the browser should remove the banner

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

## Further Work
Will this have future work?
To enable live closing of the banner, we might need to alter the Firebase action routes to pull data from Firebase and close the banner based on the information.
## Learnings
Since rebranding our application, we've been starting to use react hooks almost everywhere in our app.  This means that the way we originally used `Banner` and the `BannerContext` would no longer work in non-class components. 

Although using `useContext` and `useReducer` by themselves is pretty straightforward, combining the context state with userReducer proved to be difficult.  There aren't many working examples online, and even fewer examples _with_ typescript.

The best part is that the context api is backwards compatible, so we can still use the provider/consumer methodology on some parts of the app, and the react hooks way in others.


src/legacy/Legacy.routes.tsx
```diff
@@ -1,52 +1,45 @@
-import * as React from 'react';
-import { Route, Switch } from 'react-router-dom';
+// import * as React from 'react';
```
woeltjen: This file looks all commented out; should it be deleted?

kcvan: I was thinking about deleting it in another pr. Wanted to make sure that I wasn't deleting anything that might need to be referenced later but we can always check the git tag.  Will delete here 👍

src/utils/workBannerUtility.ts
```diff
@@ -0,0 +1,26 @@
...
+    to: '/account/verification',
+  }
+
+  if (userVerified) {
+    await bannerDispatch({ type: 'close' });
+  }
+
+  if (userUnverified) {
+    await bannerDispatch({ type: 'set', payload: options });
+    await bannerDispatch({ type: 'open' });
```
jerminatorhits: doesn't this happen synchronously without `await`

jerminatorhits: or maybe like `this.setState` it might batch the dispatch changes together

kcvan: That's a good question.  Originally we used the await in case this because a race condition on whether the banner would set banner props first or open first.  I think in the background `useReducer` still uses setState, so this might still hold true.  An option we can do is wrap the dispatch calls with `new Promise` but I think that might add unnecessary bulk.  We can try to remove it for now and see if the banner every loads before the message.  Thoughts?

0xTail:
👍

src/utils/workBannerUtility.ts
jerminatorhits: ya sounds good to me 👍

src/HOCs/FirebaseProvider.tsx
```diff
@@ -11,8 +11,8 @@
 
 import * as React from 'react';
 
-import { BannerConsumerProps } from 'HOCs/BannerProvider';
-import { showAccountVerificationBanner } from 'utils/bannerUtility';
+import { WorkBannerConsumerProps } from 'HOCs/WorkBannerProvider';
```
tc: Does this need to be prefixed `Work` if there is only one now?

kcvan: Good point! Will remove `Work` from the Banner

src/legacy/shared/Banner/Banner.tsx
```diff
@@ -39,18 +37,18 @@ const Banner = (props: Props) => {
...
-          );
-        }}
-      </AppConsumer>
+      <CloseButton
+        className="d-flex d-md-none close"
+        height="32px"
+        iconColor="upper"
+        onClose={props.onClose}
+        width="32px" />
+      <CloseButton
```
tc: Is this a copy pasta error? i see two close buttons

kcvan: When I removed the `AppConsumer` we didn't have a way to check if the user was on `mobile`, `tablet`, or `desktop` so I used two components and used `display: none` depending on the screen size (given by bootstrap)

src/utils/workBannerUtility.ts
```diff
@@ -0,0 +1,26 @@
+import { WorkBannerConsumerProps } from 'HOCs/WorkBannerProvider';
+
+export const ACCOUNT_VERIFICATION = 'Click here to verify your email and phone number to start booking!';
+
+export const showAccountVerificationBanner = (
+  isUserVerified: boolean,
+  bannerDispatch: WorkBannerConsumerProps['bannerDispatch'],
+  bannerState: WorkBannerConsumerProps['bannerState']): void => {
+  const userVerified = isUserVerified && bannerState.show;
```
tc: this can confusing: `userVerified` is equal to `isUserVerified` AND a banner state.
I don't understand why userVerified is conditional on the banner state which is a UI variable.

Did you mean `userVerified` be named something else?

kcvan: That's a good point.  We're receiving the return of `hasCompletedVerification(user)` which will tell us if a user has finished phone and email verification.  so if the user is verified and the banner is shower we want to close the banner, and the opposite if the user is unverified.  I think I can fix up the names a bit to make it more clear 🙏

src/legacy/shared/Banner/Banner.tsx
tc: ah i see; can we assign the `32px` to a variable like `mobileButtonSize` to make it more apparent you are doing this?

kcvan: Sounds good!

config/webpack.common.js
```diff
@@ -15,7 +15,6 @@ const nodeModules = path.resolve(__dirname, '..', 'node_modules');
 module.exports = {
   entry: {
     index: `${root}/index.tsx`,
-    legacy: `${root}/legacy.tsx`,
```
tc: it would have been better if these legacy code deletion were in a separate PR since they are unrelated changes.

src/utils/bannerUtility.ts
```diff
@@ -2,22 +2,23 @@ import { BannerConsumerProps } from 'HOCs/BannerProvider';
...
-  if (userVerified) {
-    await bannerActions.closeBanner();
+  if (userVerified && bannerState.show) {
+    bannerDispatch({ type: 'close' });
   }
 
-  if (userUnverified) {
-    await bannerActions.setBannerOptions(options);
-    await bannerActions.openBanner();
+  if (!(userVerified || bannerState.show)) {
```
tc: Is this more understandable? `!userVerified && !bannerState.show`
show the banner "if the user is not verified and the banner is not showing"

config/webpack.common.js
kcvan: Agreed.  Removing the original banner caused a lot of errors that eventually lead to be taking out a bunch of legacy files.  Commenting out the code seemed about the same so I decided to just remove the code.

src/utils/bannerUtility.ts
kcvan: Definitely is, fixing in next commit 🙏

PR #319, opened 2019-03-14
Add Apply button to Destination filter

woeltjen:
## Description
Updating immediately on change from this component can be disorienting for the user; instead, add an Apply button so that changes to destination and travel mode can be completed before being applied.

## How to Test
1. Search for a city
2. Open the Destination filter
3. Expect "Apply" to be visible but disabled
4. Change Destination or Travel Mode
5. Expect "Apply" to be enabled
6. Click "Apply"
7. Expect search filters to be applied

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Now that there's an Apply button, I kinda want a little reset icon next to it...

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -14,6 +14,12 @@ import GoogleMapsWithMarkersContainer from './GoogleMapsWithMarkers.container';
 
 const nearMarker = require('assets/images/iconmonstr-location-13-32.png');
 
+// https://github.com/tomchentw/react-google-maps/issues/405
+const keyFactory = {
+  counter: 0,
+  next() { return this.counter++; }
+};
+
 interface Props extends RouterProps {
   bounds?: LatLngBounds;
   children?: React.ReactNode;
@@ -114,7 +120,7 @@ class GoogleMapsWithMarkers extends React.Component<Props, State> {
           listing => !selectedListing || listing.id !== selectedListing.id
         ).map((listing, index) => (
           <OverlayView
-            key={listing.id}
+            key={keyFactory.next()}
             position={{ lat: listing.lat, lng: listing.lng }}
             mapPaneName={OverlayView.OVERLAY_MOUSE_TARGET}
           >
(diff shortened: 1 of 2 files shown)


kcvan:
works! im wondering if it's almost time to combine `Home Type` and `Destinations` in a `Filter` button, and then we can add the apply button to also work on home type

tc:
I see this error when i click "Apply"
Screen Shot 2019-03-14 at 10 10 37 AM

woeltjen:
@tc Hm, having trouble reproducing that error on console. Do you remember what place you had for Destination? Those look like errors we've seen pop up from Sentry, it'd be great to find the root cause.

woeltjen:
> works! im wondering if it's almost time to combine Home Type and Destinations in a Filter button, and then we can add the apply button to also work on home type

I like this idea! Putting this all in the same dropdown would also allow us to move it to the search bar and expose it from the front page. I'll see if I can prototype something quickly to throw on `dev02` for a UX review.

tc:
I'm using FF too.
reprod steps:
1. Search San Francisco
2. Mouse over a listing
3. Click Destination
4. Add Moscone and click Apply
5. Mouse over a listing
6. Click Destination again
7. Select Walking and Apply
I now see error

woeltjen:
Thanks! I'm able to reproduce this now.

Looks like this is tomchentw/react-google-maps#405. A solution to that (and [similar errors](https://github.com/tomchentw/react-google-maps/issues/482)) is to use unique keys on every re-render. Others have ended up needing to add a wrapper to verify that `mapPanes` is initialized in a `componentShouldUpdate` callback. I'll try the former then the latter.

woeltjen:
No longer able to reproduce the error with the fix indicated above. Sounds like `react-google-maps` is missing some initialization checks in `OverlayView`?

On the plus side, this error could probably be hit without this change (think it was introduced in #308, based on root cause and timing of Sentry errors) so we're fixing stuff!

tc:
@woeltjen with FF, i no longer get the error in console log but after clicking Apply in step 7 above, the Destination Modal doesn't close. 

I see the options updated though meaning if i selected Cycling, clicking on a hotel marker shows the cycling directions.

PR #321, opened 2019-03-14
Show rating in listing cards

woeltjen:
## Description
To support more informed booking decisions, display rating information in listing cards for search results, when that information is available.

## How to Test
1. Search for listings somewhere with Agoda listings
2. Expect to see rating information (e.g. "8.1/10 rating") in listing cards.

Note that this depends upon thebeetoken/beenest-backend#811

Which devices did you test on?
- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:
Check against these principles:

### High level
Does this code need to be written?
What are the alternatives?
Will this implementation become a support issue?
How much error margin does this solution have?

### Code
* Does the code follow industry standards?
JS: https://github.com/airbnb/javascript
React: https://github.com/airbnb/javascript/tree/master/react
https://github.com/vasanthk/react-bits
Documentation headers: http://usejsdoc.org/index.html

* Is there duplicated code? Can it be refactored into a shared method?
* Is the code consistent with our project?
* Are there unit tests? Do they test the states?
* Is the person refactoring another developer's code? If possible, did the original developer approve?

### Variables/Naming:
* Would the variable type led to future edge cases?
* Are the variable naming clear? Would the value contain something other than what the name describes.

### Security
* Can this be hacked or abused by the user?

## Further Work
* Display as stars? (Of course we're going to want to display these as stars)
* Display on listing page

PR #322, opened 2019-03-14
Display ratings in ListingCard as stars

woeltjen:
## Description
Follow up to #321: Display ratings in ListingCard as stars!

image

## How to Test
1. Search for listings
2. Expect Agoda listings to show a star rating next to "Hotel Room" in listing cards

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Show same information in ListingInformation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,6 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
-/typings
 
 # e2e
 /e2e/*.js
--- a/src/pages/search/ListingCard.tsx
+++ b/src/pages/search/ListingCard.tsx
@@ -7,6 +7,7 @@ const ListingCard = ({
   homeType,
   listingPicUrl,
   pricePerNightUsd,
+  rating,
   title
 }: ListingShort) => (
   <Card tag={Fade} className="w-100 h-100 shadow border-0">
@@ -16,7 +17,10 @@ const ListingCard = ({
       </div>
     </div>
     <CardBody>
-      <CardSubtitle className="small text-secondary">{homeType}</CardSubtitle>
+      <CardSubtitle className="small text-secondary">
+        {homeType}
+        {rating && <span> &middot; {rating.average}/10 rating</span>}
+      </CardSubtitle>
       <CardTitle tag="h6">{title}</CardTitle>
     </CardBody>
     <CardFooter>
--- a/src/pages/search/ListingCard.tsx
+++ b/src/pages/search/ListingCard.tsx
@@ -1,8 +1,6 @@
 import * as React from 'react';
 import { Card, CardBody, CardFooter, CardImg, CardSubtitle, CardText, CardTitle, Fade } from 'reactstrap';
 
-// TODO: Add a .d.ts for this, or PR one to definitelytyped?
-// @ts-ignore
 import StarRatings from 'react-star-ratings';
 
 import { ListingShort } from 'networking/listings';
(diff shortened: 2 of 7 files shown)


src/pages/search/ListingCard.tsx
```diff
@@ -1,12 +1,17 @@
 import * as React from 'react';
 import { Card, CardBody, CardFooter, CardImg, CardSubtitle, CardText, CardTitle, Fade } from 'reactstrap';
 
+// TODO: Add a .d.ts for this, or PR one to definitelytyped?
+// @ts-ignore
+import StarRatings from 'react-star-ratings';
```
woeltjen: `react-star-ratings` doesn't have types, which means [`noImplicitAny` will raise an error here](https://github.com/Microsoft/TypeScript/issues/13348#issuecomment-405361558) unless we [add and maintain a `.d.ts`](http://www.typescriptlang.org/docs/handbook/modules.html#shorthand-ambient-modules) with declarations for it.

Doing a `@ts-ignore` seems preferable to me, given the usage here, but I'm glad to go the `.d.ts` route (either here or in a subsequent PR) if anyone would prefer that

src/pages/search/ListingCard.tsx
```diff
@@ -16,7 +21,17 @@ const ListingCard = ({
       </div>
     </div>
     <CardBody>
-      <CardSubtitle className="small text-secondary">{homeType}</CardSubtitle>
+      <CardSubtitle className="small text-secondary justify-content-between row no-gutters">
+        {homeType}
+        {rating && <div style={{ marginTop: '-0.375rem' }}>
```
woeltjen: Inline style tag here to fine-tune the vertical alignment of `StarRatings`. Does anyone see a way to get similar alignment with Bootstrap classes? (`mt-n1` and `mt-n2` go too low and too high, respectively). Alternately, should this be moved to `customStyles`?

woeltjen:
Note that, at some screen widths, the stars pop down below home type:

image

That still looks reasonable to me, but figure it's worth calling out here (we could also finesse the breakpoints in the search page so that these cards don't get so narrow)

src/pages/search/ListingCard.tsx
kcvan: i think keeping the styles inline is fine, especially if it's just one style.  I've tried messing around with it but since the starts are absolutely positioned i dont think there's much we can do here.  solid fix 👍

src/pages/search/ListingCard.tsx
kcvan: is it possible to declare a new module in the d.ts and make it `content: any`? in case we use the starts elsewhere (maybe listing page?)

woeltjen: Adding custom typings (and a `typings` directory for any future cases we need to do this). It was a bit tricky getting all parts of the configuration right, but at least it ended up being a small(-ish) change set.

Went ahead and defined the Props interface for `react-star-ratings`; found it to be a useful exercise, and if we're going to expend effort satisfying TypeScript, might as well get the benefits!

.gitignore
```diff
@@ -37,7 +37,6 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
-/typings
```
woeltjen: Not sure why this was ignored, so calling attention to it here. Don't see any problems after removing it, and by convention this is where we'd store types that aren't available from `@types`. If we do want to ignore this, I can just rename that directory

PR #326, opened 2019-03-15
Refactor to consolidate search criteria

woeltjen:
## Description
To facilitate adding destination (et al) to query params during search (which allows link-sharing, and avoids losing state between refreshes), refactor to colocate the full search parameterization in a single interface that can be managed from Search

Note that this is *just* a refactor; want to consider this changeset separately to avoid breaking functionality first, then add query string synchronization when this looks good

## How to Test
Regression test search, and search filters.

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Adding additional search criteria to query params

### DIFF
--- a/src/pages/search/SearchCriteria.ts
+++ b/src/pages/search/SearchCriteria.ts
@@ -49,8 +49,8 @@ export function queryToCriteria(queryString: string): SearchFilterCriteria {
       south: parseFloat(queryParams.bounds.south)
     },
     coordinates: queryParams.coordinates && {
-      lat: parseFloat(queryParams.bounds.lat),
-      lng: parseFloat(queryParams.bounds.lng)
+      lat: parseFloat(queryParams.coordinates.lat),
+      lng: parseFloat(queryParams.coordinates.lng)
     },
     homeType: queryParams.homeType,
     numberOfGuests: queryParams.numberOfGuests && parseInt(queryParams.numberOfGuests),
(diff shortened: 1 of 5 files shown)


src/pages/search/SearchCriteria.ts
```diff
@@ -1,17 +1,85 @@
...
+// Redundant to google.maps.TravelMode, but that may be unavailable at
+// initialization-time; it loads asynchronously.
+export enum TravelMode {
+  DRIVING,
+  WALKING,
+  TRANSIT,
+  BICYCLING
+}
 
 export interface SearchFilterCriteria {
```
woeltjen: At this point, SearchFilterCriteria is *very close* to being identical to ListingSearchInput

It's tempting to unify the two. Reasons I haven't:

* `near.name` is needed by the front-end for display purposes, but would not be used on the backend
* `travelMode` is also not used by the backend

In short, unifying these interfaces looks like it would require backend interface changes without any changes to backend functionality, which seems rather backwards. Keeping the interfaces separate leaves us with the boring `toListingSearchInput` to maintain, which just omits the properties which aren't used on the backend. That seems like the less undesirable option to me.

kcvan: i think keeping these separate is no issue. it sounds like the pros of combining them arent significant enough to justify the changes for now

PR #327, opened 2019-03-15
Keep search filters in query string

woeltjen:
## Description
Keeps user's chosen destination (and other search filter parameters, why not?) in the query string so that specific searches can easily be shared, information is not lost on reload, and so on.

Note that this has been branched from #326 and should be reviewed/merged after that PR has been merged

## How to Test
1. Make a search
2. For each of Destination, Travel Mode, and Home Type:
    1. Change the filter
    2. Reload
    3. Verify selection is still present and reflected on map and in listing results

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -7,7 +7,7 @@ import { SETTINGS } from 'configs/settings';
 const { GOOGLE_MAPS_KEY } = SETTINGS;
 
 import { ListingCard } from 'legacy/shared/ListingCard';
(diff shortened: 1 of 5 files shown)


woeltjen:
> probably of of the scope of this pr, but it would be great if we could select a destination, get the directions, and be able to send this in the url for someone else and have it show the driving directions also

Hm, interesting thought. Maybe this could be done by keeping the selected listing in the query string? Or, we could keep search filters in query params and add directions to individual listing pages

In any event, agree that this is out of scope for now

PR #328, opened 2019-03-15
Move destination marker back to front

woeltjen:
## Description
#308 introduced a subtle regression: Pricing overlays obscure the destination marker, making it difficult to find (in some circumstances) even when it is in the map's bounds.

Root cause was the switch to "overlays" (which allow custom markup) from "markers" (which are just images or paths); the former exists on a separate layer from the latter. In particular, we need to use the `OVERLAY_MOUSE_TARGET` layer to support mouse interactions (clicking on pricing markers), but that layer is in front of the layer which displays markers (and the destination is still a marker)

Solution here is hacky: Move the pricing display to `OVERLAY_LAYER`, which appears below the marker layer. Then, to support mouse interactions, include a redundant set of overlays in the `OVERLAY_MOUSE_TARGET` layer with same properties but an opacity of `0`

## How to Test
1. Do a search
2. Choose a destination near many listings (e.g. Moscone Center)
3. Expect destination marker to obscure listings (and not the other way around)

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -118,11 +118,11 @@ class GoogleMapsWithMarkers extends React.Component<Props, State> {
         />}
         {listings.filter(
           listing => !selectedListing || listing.id !== selectedListing.id
-        ).map((listing, index) => (
+        ).map((listing, index) => (<>
           <OverlayView
             key={keyFactory.next()}
             position={{ lat: listing.lat, lng: listing.lng }}
-            mapPaneName={OverlayView.OVERLAY_MOUSE_TARGET}
+            mapPaneName={OverlayView.OVERLAY_LAYER}
           >
             <button className="popover p-1 bs-popover-top" style={{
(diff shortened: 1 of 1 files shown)


kcvan:
i think there might be a missing `key={x}` on a map, getting:
```
Warning: Each child in a list should have a unique "key" prop.
```

kcvan:
also just to be sure, the destination marker should be obscuring the listing prices, not necessarily the listing pop up card, right?

woeltjen:
> also just to be sure, the destination marker should be obscuring the listing prices, not necessarily the listing pop up card, right?

Yep (think obscuring the listing pop up would look bad; think changing the opacity of the info window is the solution there)

woeltjen:
> i think there might be a missing `key={x}` on a map, getting:

Ah, thanks! There were keys there, but they were in a `<></>` which seems to throw off React for whatever reason. Switched to two loops so that both could return arrays of single, keyed Elements, and the error went away.

PR #330, opened 2019-03-15
Deploy

woeltjen:
Star ratings; banners; typos

**Depends on thebeetoken/beenest-backend#812 for rating information**


woeltjen:
Backend deployment (upon which this depends) is stalled currently while env gets sorted out; going to close this so that we don't accidentally deploy something that will break.

PR #332, opened 2019-03-16
Scroll selected listings into view

woeltjen:
## Description
When a listing is selected from the map, scroll that listing into view.

This is branched from #331 and should be reviewed/merged after that PR

## How to Test
1. Search!
2. Click on a listing's marker on the map
3. Expect to see the related listing scroll into view on the left
4. Hover over listings on the left
5. Expect those listings to select, but to stay put

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
This makes me wonder if showing listing cards on map is something we need to keep doing at all. It would solve the "listing info obscures driving directions" problem we run into some times

## Learnings
`scrollIntoView` doesn't check if an element is already in view

### DIFF
--- a/src/pages/search/SearchResults.tsx
+++ b/src/pages/search/SearchResults.tsx
@@ -19,6 +19,11 @@ interface Props extends Params {
   selectedListing?: ListingShort;
 }
 
+// Link.innerRef doesn't accept a React ref as a propType, so fake it
+interface FakeRef {
+  current: Element | null;
+}
+
 const listingPath = (listing: ListingShort, params: Params) =>
   `/listings/${listing.idSlug}?${stringifyQueryString(params)}`;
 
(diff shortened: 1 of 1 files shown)


woeltjen:
Hm, was that in [Safari](https://caniuse.com/#search=scrollIntoView), by any chance? I'm passing `block: 'center'` in, which should scroll the element to vertical center. Looks like those parameters are unsupported in Safari/IE/Edge, so it just ends up at the top.

The reason it's partially obscured is because of the header. To address that, we could refactor the header to be neither `fixed` nor `sticky`, and to instead use a scrollable container below it to hold the rest of the page. Might address some of our other issues with the header too?

On the other hand, I'd be comfortable living with that particular issue; it's still an improvement to leaving those listings off-page!

kcvan:
was on chrome when it happened! (who browses in safari😝)  im also good leaving it as is, LGTM!

PR #337, opened 2019-03-18
Add a Host CTA to home page

woeltjen:
## Description
Adds a CTA to solicit listings to the home page, similar to what we had prior to the redesign

## How to Test
1. Go to the home page
2. Click "I'm a Host"
3. Expect an "Add a Listing" button to be visible
4. Click that button
5. Expect to go to `/host/listings` (if signed in; otherwise, to signup form)

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Do we want to dynamically decide the functionality of this button based on the current user, like we do in the Header? (I started down that path, then decided to start simple.)

### DIFF
--- a/src/pages/home/Home.tsx
+++ b/src/pages/home/Home.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { Button, Col, Container, Fade, Jumbotron, Row } from 'reactstrap';
+import { Link } from 'react-router-dom';
 
 import { affiliations, guestValueProps, HomeUser, hostValueProps, testimonials } from './home.config';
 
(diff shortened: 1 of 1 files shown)


jerminatorhits:
looks good. For the mobile view, I think buttons are typically full width

woeltjen:
> looks good. For the mobile view, I think buttons are typically full width

Ah, thanks! I've seen that often but haven't picked up on it as a design principle.

Increased width of the button on mobile. At first I tried to get it to match the widths of the host/guest switch buttons above, but the widths of the containing elements are different enough that I had trouble with that. Ended up going with a `w-100` with some padding that looked reasonable:

image

PR #338, opened 2019-03-19
Reduce info window opacity when driving directions show

woeltjen:
## Description
To ensure driving directions are not obscured by the Info Window containing listing information, reduce the opacity of that window when directions are being shown (and restore opacity on hover)

## How to Test
1. Do a search which results in some listings
2. Open Destination filter, choose a place and travel mode, click Apply
3. Hover over listings on the left
4. Expect Info Window to appear at 75% opacity, with driving directions visible underneath (in those cases when they'd be otherwise obscured)
5. Click on listing in the map to open its Info Window, expect that to be at 75%
6. Hover over that Info Window
7. Expect full opacity at that point
8. Clear Destination, click Apply
9. Hover over listings on left
10. Expect listing Info Windows in map to be full opacity

image

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/styled/customStyles.scss
+++ b/src/styled/customStyles.scss
@@ -122,9 +122,11 @@ $HEADER-HEIGHT: 4.125rem;
 .bee-directions-showing {
   .gm-style-iw {
     opacity: 0.75;
+    transition: all 0.2s ease-in-out;
   }
 
   .gm-style .gm-style-iw:hover {
     opacity: 1;
+    transition: all 0.2s ease-in-out;
   }
 }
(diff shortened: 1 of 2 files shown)


src/styled/customStyles.scss
```diff
@@ -117,4 +117,14 @@ $HEADER-HEIGHT: 4.125rem;
 #bee-modal-root {
   position: relative;
   z-index: 10000;
+}
+
+.bee-directions-showing {
+  .gm-style-iw {
+    opacity: 0.75;
```
jerminatorhits: we could add `transition: all 0.2s ease-in-out;` to smoothen transition

woeltjen: Nice touch! Added

PR #339, opened 2019-03-19
Show route duration on map WIP

woeltjen:
## Description
To provide users with more useful context when selecting hotels, show duration associated with driving directions on map

## How to Test
1. Search
2. Enter a Destination filter
3. Select listings to show directions
4. Expect duration of travel to display on map, half-way along the path

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
+++ b/src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
@@ -8,7 +8,7 @@ const { GOOGLE_MAPS_KEY } = SETTINGS;
 
 import { ListingCard } from 'legacy/shared/ListingCard';
 import { LatLng, LatLngBounds, ListingShort } from 'networking/listings';
-import { formatPriceShort } from 'utils/formatter';
+import { formatPriceShort, formatDuration } from 'utils/formatter';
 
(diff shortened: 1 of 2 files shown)


src/legacy/shared/GoogleMapsWithMarkers/GoogleMapsWithMarkers.tsx
```diff
@@ -208,3 +220,31 @@ function getCenterCoordinates(listings: ListingShort[]) {
...
+    travelled += distances[start];
+    start += 1;
+  }
+  const remaining = total - travelled;
+  const endWeight = remaining / distances[start];
+  const startWeight = 1 - endWeight;
+  return {
+    lat: startWeight * path[start].lat() + endWeight * path[start + 1].lat(),
+    lng: startWeight * path[start].lng() + endWeight * path[start + 1].lng()
+  };
```
woeltjen: This function is currently pretty darn broken; unfortunately, there's not a "center of this route" property on the directions result.

woeltjen:
This is a nice-to-have and level-of-effort looks non-trivial; closing for now

PR #344, opened 2019-03-20
Add Avatar to Header

woeltjen:
## Description
To provide users with situational awareness (am I logged in as who I think I am?) show Avatar in the top-right of the header

## How to Test
1. Log in as a user with a configured profile picture
2. Expect to see that profile picture in the top right
3. Click on that picture
4. Expect to go to `/account/general`

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

### DIFF
--- a/src/legacy/work/Header/Header.tsx
+++ b/src/legacy/work/Header/Header.tsx
@@ -35,6 +35,7 @@ const authNavItems = [
   {
     header: 'Account',
     link: '/account',
+    className: 'd-md-none'
   },
   {
     header: 'Trips',
@@ -107,7 +108,7 @@ const DetailedHeader = () => {
                               {isHost ? 'Host Profile' : 'Become a Host'}
                             </Link>
                             {authNavItems.map(item => (
-                              <NavItem className="px-2" key={item.header}>
+                              <NavItem className={['px-2', item.className].join(' ')} key={item.header}>
                                 <NavLink
                                   onClick={() => toggleNavbar(false)}
                                   tag={Link}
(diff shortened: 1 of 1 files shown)


jerminatorhits:
works as intended. could we remove the "Account" tab for desktop since clicking the avatar will now route to 'account'

src/legacy/work/Header/Header.tsx
```diff
@@ -106,7 +108,7 @@ const DetailedHeader = () => {
                               {isHost ? 'Host Profile' : 'Become a Host'}
                             </Link>
                             {authNavItems.map(item => (
-                              <NavItem className="px-2" key={item.header}>
+                              <NavItem className={['px-2', item.className].join(' ')} key={item.header}>
```
woeltjen: This surprised me a little bit:

    $ node
    > ['a', undefined].join(' ')
    'a '

I'd have expected `'a undefined'`

We tend to do a bit more finessing when `className` comes from multiple places, so figured I'd call this out in case there is something I'm missing.

tc: there's a package for that :) https://www.npmjs.com/package/classnames

woeltjen: Hehe, yeah, ran across that in [this thread](https://github.com/facebook/react/issues/3138), which is also where the [`join`](https://github.com/facebook/react/issues/3138#issuecomment-74311220) solution came from. Almost went with the former but couldn't justify it when a built-in is functionally equivalent (for this usage).

Only drawback I see with `join` is that the behavior is potentially non-obvious, at the language level:

    ['a', undefined, 'b'].join(' ') !== 'a' + ' ' + undefined + ' ' + 'b'

PR #345, opened 2019-03-20
Fix bug with Destination filter

woeltjen:
## Description
Pass both filter parameters up at the same time on apply, to avoid clobbering changes from separate callbacks.

This addresses the issue noted in Further Work of #341. Changing both destination and travel mode with separate calls on the same Apply caused the former to be lost, because of the nature of the handlers involved.

## How to Test
1. Search (e.g. for San Francisco)
2. Open the Destination dropdown
3. Enter a place (e.g. Coit Tower; select it from the list)
4. Change the travel mode
5. Click Apply
6. Expect destination to appear on map and for travel mode preference to be retained/reflected

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
None expected!

## Learnings
`useState` has some "quirks" (at least if you come at it expecting it to work like `setState`)

In particular, this two-callback solution:

    onPlaceChange={near => onFilterChange({ ...filter, near })}
    onTravelModeChange={travelMode => onFilterChange({ ...filter, travelMode })}

...caused the loss of `near` when the two are executed in sequence, because `filter` is still the same value returned from the prior `useState` call in both cases.

### DIFF
--- a/src/pages/search/SearchForm.tsx
+++ b/src/pages/search/SearchForm.tsx
@@ -25,8 +25,7 @@ const SearchForm = ({ filter, onFilterChange }: Props) => {
           <TransitTime
             place={filter.near}
             travelMode={filter.travelMode}
-            onPlaceChange={near => onFilterChange({ ...filter, near })}
-            onTravelModeChange={travelMode => onFilterChange({ ...filter, travelMode })}
+            onChange={(near, travelMode) => onFilterChange({ ...filter, near, travelMode })}
           />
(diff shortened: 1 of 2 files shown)


kcvan:
works as expected!

PR #353, opened 2019-03-25
delete old Admin and Hosts

kcvan:
## Description
Why did you write this code?
cleaning up old code that isnt used.  also transferred google maps components from legacy to shared so both folders are the same
## How to Test
- [ ] Visit site/admin/host sign up
- [ ] Verify site/admin/host sign up works

- [ ] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings


src/legacy/shared/GoogleMaps/GoogleMaps.tsx
```diff
@@ -8,8 +8,6 @@ const { GOOGLE_MAPS_KEY } = SETTINGS;
 import GoogleMapsContainer from './GoogleMaps.container';
 import AudioLoading from 'legacy/shared/loading/AudioLoading';
 
-import { useDebounce } from 'utils/hooks';
```
woeltjen: This looks like it unintentionally reverts a change from #292

tc: @kcvan bump

kcvan: ah good catch, let me actually close this PR and make a smaller one for just deleting admin and host

PR #369, opened 2019-03-28
direct search error page to correct page

kcvan:
## Description
Why did you write this code?
search error page would not redirect by url params change when directing to the same base url, ex.
```
/search
```
to
```
/search?locationQuery=San%20Francisco%2C%20CA%2C%20USA
```

would not reload the page because we're effectively on the same route.
 
## How to Test
- [ ] On the home page, click search without any parameters
- [ ] Enter a city from the error page, it should take you to the correct city
- [ ] Test search/listing page normally

- [x] Chrome on Mac
- [ ] Chrome on PC
- [ ] Firefox on Mac
- [ ] Firefox on PC
- [ ] Safari iPhone
- [ ] Chrome Android

## REVIEWERS:

### High level
What are the alternatives?

### Code

### Variables/Naming:

### Security

## Further Work
Will this have future work?

## Learnings

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -46,7 +46,7 @@ class ScrollToTop extends React.Component<RouterProps, {}> {
 
 const ScrollToTopWithRouter = withRouter(ScrollToTop);
 
-const Work = () => {
+const Work = (props: RouterProps) => {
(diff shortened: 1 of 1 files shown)


jerminatorhits (approved):
lgtm. empty search page is now operational

